PR #356, opened 2025-04-03
Replaced nac-test with iac-test as package name in Readme.md

krishna1306:
`nac-test` is not a valid python package. `iac-test` is. Hence, replaced all references to `nac-test` with `iac-test`


danischm:
We are in the process of migrating to the new name/package. I will publish a pre-release later today, though you can still continue using `iac-test`.

PR #390, opened 2025-10-15
Add support for test case parallelization for capable suites

oboehmer:
This PR makes the following changes:

If the suite's metadata contains `Test Concurrency` with a True value, `nac-test` will schedule the tests in this suite to be run in parallel. We are using an orderingfile plus `--testlevelsplit` to achieve this.

### DIFF
--- a/nac_test/cli/main.py
+++ b/nac_test/cli/main.py
@@ -213,11 +213,12 @@ def main(
     """A CLI tool to render and execute Robot Framework tests using Jinja templating."""
     configure_logging(verbosity)
 
+    if "NAC_TEST_NO_TESTLEVELSPLIT" not in os.environ:
+        ordering_file = output / ORDERING_FILE
+    else:
+        ordering_file = None
+
     try:
-        if "NAC_TEST_NO_TESTLEVELSPLIT" not in os.environ:
-            ordering_file = output / ORDERING_FILE
-        else:
-            ordering_file = None
         writer = nac_test.robot_writer.RobotWriter(
             data, filters, tests, include, exclude
         )


nac_test/cli/main.py
```diff
@@ -21,7 +21,7 @@ def configure_logging(level: str) -> None:
         lev = logging.DEBUG
     elif level == "INFO":
         lev = logging.INFO
-    elif level == "WARNING":
+    elif level == "WARN":
```
danischm: Shouldn't this remain "WARNING"?

nac_test/cli/main.py
```diff
@@ -21,7 +21,7 @@ def configure_logging(level: str) -> None:
...
         lev = logging.WARNING
     elif level == "ERROR":
```
danischm: If we introduce "TRACE" I think we should map it here as well.

nac_test/pabot.py
```diff
@@ -1,22 +1,46 @@
...
 
+import os
+import re
 from pathlib import Path
 
 import pabot.pabot
 
+ORDERING_FILE = "ordering.txt"
+
+ROBOT_LOGLEVEL_DICT = {
```
danischm: Should we map all log levels here or rather rely on the default value?

nac_test/robot_writer.py
```diff
@@ -118,7 +143,71 @@ def _fix_duplicate_path(self, *paths: str) -> Path:
...
+            Input:  output_path = foobar
+                    filename = foobar/integration_tests/whatever.robot
+            Result: suite_dirname = Foobar.Integration Tests.Whatever
+        """
+        relative_path = robot_file.parent.relative_to(output_path)
+        path_parts = [output_path.name] + list(relative_path.parts) + [robot_file.stem]
+        return ".".join([TestSuite.name_from_source(p) for p in path_parts if p])
+
+    def _update_ordering_entries(
+        self, output_path: Path, robot_file: Path, delete_empty_suite: bool = False
```
danischm: Do we need "delete_empty_suite" if we always set it to "true"?

danischm (changes requested):

oboehmer (commented):
Will review the loglevel stuff..

nac_test/robot_writer.py
oboehmer: I just pushed a diff to remove this logic.. too dangerous if you ask me..

nac_test/robot_writer.py
```diff
@@ -199,8 +279,24 @@ def write(self, templates_path: Path, output_path: Path) -> None:
...
                 o_path = Path(output_path, rel, filename)
                 self.render_template(t_path, o_path, env)
+                self._update_ordering_entries(output_path, o_path)
+
+        if (
+            ordering_file
+            and len(self.ordering_entries) > 0
+            # only create ordering file if there is a need for test concurrency
+            and any(o.startswith("--test") for o in self.ordering_entries)
+        ):
```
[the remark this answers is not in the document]
oboehmer: good point, also triggered me to put control of ordering_file creation and processing to `main()`

danischm (approved):

PR #529, opened 2026-02-10
fix(pyats): add disconnect/wait timeout configuration to testbed generator

### LINKED ISSUE #527: Device disconnection takes 11 secs * number of devices, need to optimize
## Problem

After disconnecting from devices, we observe a **10-second delay** before the next disconnect proceeds. This delay is visible in timestamped logs and adds unnecessary overhead to test execution.

## Evidence

Timestamps show consistent 11-second intervals between sequential disconnects:

```
2026-02-09 16:30:02,934 - INFO - Disconnected from device: sd-dc-c8kv-01
2026-02-09 16:30:13,954 - INFO - Disconnected from device: sd-dc-c8kv-03  (+11s)
2026-02-09 16:30:24,978 - INFO - Disconnected from device: sd-dc-c8kv-02  (+11s)
2026-02-09 16:30:36,000 - INFO - Disconnected from device: sd-dc-c8kv-04  (+11s)
```

This indicates a 10-second delay between each disconnect operation, plus ~1 second for the disconnect itself.

## Root Cause

Unicon has a built-in `POST_DISCONNECT_WAIT_SEC = 10` setting (cooldown period after disconnect) defined in `unicon/src/unicon/settings.py`. This prevents "Connection Refused" errors during rapid disconnect/reconnect cycles.

**However:** When using ConnectionBroker, this delay is unnecessary since the broker manages connection lifecycle and doesn't do back-to-back disconnect/reconnect.

## Initial Approach (Ineffective)

Attempted to override the setting in testbed connection configuration:

```python
# nac_test/pyats_core/execution/device/testbed_generator.py
connection_args = {
    "protocol": "ssh",
    "ip": device["host"],
    "port": device.get("port", 22),
    "arguments": {...},
    "settings": {
        "POST_DISCONNECT_WAIT_SEC": 0,  # Attempted override - did not work
    },
}
```

**Result:** The 10-second delay still occurs. This approach does not effectively override the Unicon setting.

## Impact

- **Current:** 4 devices × 10s delay = 40 seconds overhead per test run
- **Scale:** Grows linearly with device count (20 devices = 3m 20s)
- **Severity:** Minor compared to Task().run() optimization, but worth fixing
- **Occurrence:** Once per device after all D2D tests complete

## Next Steps

**Requires further investigation:**

1. Determine correct method to override Unicon settings in PyATS testbed
2. Investigate alternative approaches:
   - Global settings override before ConnectionBroker initialization
   - Runtime override when broker connects to devices
   - Explicit setting in ConnectionBroker disconnect logic
3. Verify if setting needs to be applied at connection object level vs testbed level
4. Test if override works and doesn't cause connection errors

## References

- Unicon settings: `unicon/src/unicon/settings.py` (line 18)
- Disconnect implementation: `unicon/src/unicon/eal/backend/pty_backend.py` (close method)
- Analysis: `.sisyphus/drafts/unicon-disconnect-delay-analysis.md`

oboehmer:
## Summary

Adds support for configuring disconnect and wait timeout parameters in the PyATS testbed generator, enabling better control over connection lifecycle management during test execution.

this change also improves test terardown time greatly improving execution times for d2d (not sure why, we keep the session established between tests).

Fixes #527

## Changes

- **testbed_generator.py**: Extended testbed configuration to include `disconnect_timeout` and `wait_timeout` parameters
- **test_testbed_generator.py**: Added unit tests for baseline timeout configuration
- **test_testbed_generator_merge.py**: Added comprehensive tests for merged configurations (18 new test cases)

## Testing

- ✅ Unit tests for baseline timeout configuration added
- ✅ Comprehensive merge configuration tests (18+ test cases)
- ✅ Backward compatibility maintained - existing configurations unaffected
- ✅ Both disconnect and wait timeout parameters tested in isolation and merged contexts

## Checklist

- [x] Tests pass locally
- [x] Changes are focused and atomic
- [x] Commit history is clean
- [x] Code follows project conventions

### DIFF
--- a/tests/e2e/fixtures/pyats_api_only/data.yaml
+++ b/tests/e2e/fixtures/pyats_api_only/data.yaml
@@ -1,4 +1,8 @@
+# ACI e2e tests need both top-level `apic` and `default` dicts defined
+# at runtime (actual dict content doesn't matter for these tests),
+# define them here in a common file.
 apic:
+  version: 6.0
 defaults:
   apic:
     version: 6.0


oboehmer:
thanks for the review, @aitestino .. comments addressed.

aitestino (changes requested):
Hey @oboehmer,

Thanks for addressing the feedback from the previous review—the constant extraction and test helper refactoring both look great. Just a couple of things that stood out during this round that might be worth tidying up:

### 1. DRY Opportunity in Production Code

I noticed the `arguments` dict is duplicated in two conditional branches in `_build_device_config()` (around lines 247-250 and 257-260):

```python
# Branch 1
connection_args = {
    ...
    "arguments": {
        "init_config_commands": [],
        "operating_mode": True,
    },
    ...
}

# Branch 2 - identical arguments dict
connection_args = {
    ...
    "arguments": {
        "init_config_commands": [],
        "operating_mode": True,
    },
    ...
}
```

If we ever add another optimization to `arguments`, we'd need to update both places. Could we extract this to a helper function or define it once before the conditional logic? Something like:

```python
default_arguments = {
    "init_config_commands": [],
    "operating_mode": True,
}

# Then use it in both branches
connection_args = {
    ...
    "arguments": default_arguments,
    ...
}
```

This way, future changes to the arguments structure only need to happen in one spot.

---

### 2. Test Helper Could Import the Constant

In `assert_connection_has_optimizations()`, the assertion hardcodes `0` instead of using the constant:

```python
assert connection["settings"]["POST_DISCONNECT_WAIT_SEC"] == 0  # <-- Hardcoded
```

Would it make sense to import `PYATS_POST_DISCONNECT_WAIT_SEC` and use that instead? That way, if we ever change the constant value, the test automatically validates against the new value without needing updates:

```python
from nac_test.pyats_core.constants import PYATS_POST_DISCONNECT_WAIT_SEC

def assert_connection_has_optimizations(connection: dict[str, Any]) -> None:
    ...
    assert connection["settings"]["POST_DISCONNECT_WAIT_SEC"] == PYATS_POST_DISCONNECT_WAIT_SEC
```

---

### 3. Naming Consistency (Minor)

The constant uses `SEC` as an abbreviation:

```python
PYATS_POST_DISCONNECT_WAIT_SEC: int = 0
```

Python style guides generally favor full words for clarity. Could we rename this to `PYATS_POST_DISCONNECT_WAIT_SECONDS` to be maximally clear? It's a bit longer, but it removes any ambiguity about the abbreviation.

---

These are all pretty minor—the core implementation is solid and correctly addresses the performance issue. Let me know if any of this doesn't make sense or if you'd like to discuss the tradeoffs!

---

PR #535, opened 2026-02-16
chore: add license-headers check to pre-commit hooks

oboehmer:
- added license check to pre-commit hooks (I keep forgetting for new files ;-) )
- Accept optional file arguments for pre-commit integration
- Standalone mode checks all tracked Python files (replaced `find` with `git ls-files` to respect .gitignore)


danischm:
Would be great if you could open the equivalent PR for nac-validate as well.

Issue #541, opened 2026-02-18
Consolidate `tests/unit` and `tests/pyats_core` to simplify test organization


oboehmer:
## Background

As the codebase has grown, we now have two directories (`tests/unit` and `tests/pyats_core`) containing unit-style tests. This analysis examines the current structure and proposes options for simplifying test organization going forward.

---

## Current Test Directory Structure

```
tests/
├── conftest.py              # Global fixtures (mock server, env cleanup)
├── unit/                    # 19 test files
│   ├── cli/
│   │   └── test_diagnostic.py
│   ├── core/
│   │   ├── test_combined_generator.py
│   │   └── test_types.py
│   ├── pyats_core/
│   │   └── execution/
│   │       └── test_subprocess_runner.py
│   ├── robot/
│   │   ├── test_robot_generator.py
│   │   ├── test_orchestrator.py
│   │   ├── test_robot_output_parser.py
│   │   └── test_pabot_error_handling.py
│   └── [11 more test files at root]
├── pyats_core/              # 8 test files
│   ├── common/
│   │   ├── test_auth_cache.py
│   │   ├── test_base_test_controller_detection.py
│   │   ├── test_base_test_result_collector.py
│   │   ├── test_subprocess_auth.py
│   │   └── test_subprocess_http_client.py
│   ├── execution/
│   │   └── test_subprocess_runner_integration.py
│   ├── test_orchestrator_controller_detection.py
│   └── test_orchestrator_controller_param.py
├── integration/             # 7 test files
└── e2e/                     # 1 test file + fixtures
```

---

### Test Patterns

Both `tests/unit/` and `tests/pyats_core/` use the same testing approach:
- pytest fixtures
- `unittest.mock` (patch, MagicMock, AsyncMock)
- `monkeypatch` for environment variables
- Class-based test organization (predominantly)

### Directory Contents

| Directory | Test Count | Tests For |
|-----------|------------|-----------|
| `tests/unit/` | 19 files | cli, core, robot, pyats_core modules |
| `tests/pyats_core/` | 8 files | pyats_core module |
| `tests/integration/` | 7 files | CLI integration scenarios |
| `tests/e2e/` | 1 file | End-to-end workflows |

### Observations

1. Both `tests/unit/pyats_core/` and `tests/pyats_core/` contain tests for the `nac_test.pyats_core` module
2. The `test_subprocess_runner*.py` files in both locations test complementary aspects of the same class
[...]

PR #554, opened 2026-02-19
feat(dry-run): add PyATS dry-run support

### LINKED ISSUE #460: [bug] `--dry-run` option is ignored for pyats tests
pyats test orchestrator currently ignores the `--dry-run` command option and executes discovered test cases. 
We need to define what we want to execute when dry-run is active, but we definitely should not execute tests :) 

`--render-only` also needs to be discussed in this context
Update: render-only will be addressed in #509

### LINKED ISSUE #579: E2EScenario.validate() method is never invoked
## Issue

`E2EScenario.validate()` contains robust D2D hostname validation logic but is never called anywhere in the codebase.

**File**: `tests/e2e/config.py:72-79`

Four test methods contain comments claiming "Guaranteed by E2EScenario.validate()" but this is false since the method is never invoked:
- `tests/e2e/test_e2e_scenarios.py:270`
- `tests/e2e/test_e2e_scenarios.py:287`
- `tests/e2e/test_e2e_scenarios.py:304`
- `tests/e2e/test_e2e_scenarios.py:337`

## Fix

Call `scenario.validate()` in test setup (e.g., `conftest.py` fixture or `__post_init__`).

## Files
- `tests/e2e/config.py:72-79`
- `tests/e2e/test_e2e_scenarios.py:270,287,304,337`

oboehmer:
Fixes #460
Fixes #579

## Description

Add `--dry-run` support for PyATS tests. Previously, `--dry-run` only worked for Robot Framework tests while PyATS tests would execute anyway.

**PyATS dry-run behavior:**
- Discovers and categorizes tests (API vs D2D) normally
- Prints which tests would be executed without actually running them
- Returns exit code 0 (no tests executed = no failures)

Example output:
```
======================================================================
🔍 DRY-RUN MODE: Showing tests that would be executed
======================================================================

📋 API Tests (2):
   • tests/verify_sdwan_sync.py
   • tests/verify_aci_tenant.py

📋 D2D/SSH Tests (1):
   • tests/d2d/verify_iosxe_control.py

======================================================================
✅ PyATS dry-run complete (no tests executed)
======================================================================
```

## Closes

- Fixes #460
- Fixes #579

## Related Issue(s)

- #509 (render-only to be addressed separately)

## Type of Change

- [ ] Refactoring / Technical debt (internal improvements with no user-facing changes)
- [ ] Documentation update

## Test Framework Affected

- [x] PyATS
- [x] Robot Framework
- [ ] Both

- [ ] ACI (APIC)
- [ ] Meraki (Cloud-managed)
- [ ] Hyperfabric
- [x] All architectures

## Platform Tested

- [x] macOS (version tested: macOS with Python 3.12)

## Key Changes

- Add `dry_run` parameter to `PyATSOrchestrator.__init__()`
- Add `_print_dry_run_summary()` method to display discovered tests
- Update `CombinedOrchestrator` to pass `dry_run` to `PyATSOrchestrator`
- Update CLI `--dry-run` help text to describe PyATS behavior
- Add 6 unit tests for PyATS dry-run functionality
- Add 11 E2E tests with mixed Robot + PyATS scenario

## Testing Done

  - [x] D2D/SSH tests executed successfully (if applicable)

### Test Commands Used

```bash
# Run all tests
uv run pytest -n auto --dist loadscope
# 658 passed, 71 skipped

# Run dry-run specific tests
uv run pytest tests/pyats_core/test_orchestrator_dry_run.py tests/e2e/test_e2e_scenarios.py::TestE2EDryRun -v
# 17 passed

# Pre-commit checks
uv run pre-commit run --all-files
# All passed
```

## Checklist

N/A

## Additional Notes

The dry-run implementation intercepts after test discovery and categorization but before job file generation and execution. This ensures that all validation (file discovery, categorization logic) is exercised while preventing actual test execution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,7 @@ pabot_results/
 CLAUDE.md
 # Opencode/oh-my-opencode
 .sisyphus/
+AGENTS.md
 
 # bandit security report, if generated
 bandit-security-report.json
--- a/nac_test/combined_orchestrator.py
+++ b/nac_test/combined_orchestrator.py
@@ -175,9 +175,9 @@ class CombinedOrchestrator:
 
         # Build combined results from individual orchestrators
         combined_results = CombinedResults()
+        mode_suffix = " (dry-run)" if self.dry_run else ""
 
         if has_pyats and not self.render_only:
-            mode_suffix = " (dry-run)" if self.dry_run else ""
             typer.echo(f"\n🧪 Running PyATS tests{mode_suffix}...\n")
             self._check_python_version()
 
@@ -200,7 +200,6 @@ class CombinedOrchestrator:
             combined_results.d2d = pyats_results.d2d
 
         if has_robot:
-            mode_suffix = " (dry-run)" if self.dry_run else ""
             typer.echo(f"\n🤖 Running Robot Framework tests{mode_suffix}...\n")
 
             robot_orchestrator = RobotOrchestrator(
--- a/nac_test/core/constants.py
+++ b/nac_test/core/constants.py
@@ -38,6 +38,9 @@ EXIT_DATA_ERROR = 252  # Invalid Robot Framework arguments OR no tests found (ma
 EXIT_INTERRUPTED = 253  # Execution was interrupted (Ctrl+C, etc.)
 EXIT_ERROR = 255  # Infrastructure/execution errors occurred
 
+# Reason string used in TestResults.not_run() for dry-run
+DRY_RUN_REASON = "dry-run mode"
+
 # Output directory structure - single source of truth for directory layout
 # These define the standardized paths for test results and reports
 PYATS_RESULTS_DIRNAME = "pyats_results"
--- a/nac_test/pyats_core/orchestrator.py
+++ b/nac_test/pyats_core/orchestrator.py
@@ -16,6 +16,7 @@ from typing import Any
 import yaml
 
 from nac_test.core.constants import (
+    DRY_RUN_REASON,
     EXIT_ERROR,
     PYATS_RESULTS_DIRNAME,
     SUMMARY_REPORT_FILENAME,
@@ -535,9 +536,6 @@ class PyATSOrchestrator:
         Args:
             api_tests: List of discovered API test files
             d2d_tests: List of discovered D2D test files
-
-        Returns:
-            None
         """
         print("\n" + "=" * 70)
         print("🔍 DRY-RUN MODE: Showing tests that would be executed")
@@ -616,8 +614,8 @@ class PyATSOrchestrator:
         # Dry-run mode: print discovered tests and return results without further execution
         if self.dry_run:
             self._print_dry_run_summary(api_tests, d2d_tests)
-            api_result = TestResults.not_run("dry-run mode") if api_tests else None
-            d2d_result = TestResults.not_run("dry-run mode") if d2d_tests else None
+            api_result = TestResults.not_run(DRY_RUN_REASON) if api_tests else None
+            d2d_result = TestResults.not_run(DRY_RUN_REASON) if d2d_tests else None
             return PyATSResults(api=api_result, d2d=d2d_result)
 
         breakdown_parts = []
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -22,7 +22,12 @@ import pytest
 from typer.testing import CliRunner
 
 import nac_test.cli.main
-from tests.e2e.config import E2EScenario
+from tests.e2e.config import (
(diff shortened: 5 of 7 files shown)


aitestino (changes requested):
Hey @oboehmer, nice work on this — the intercept point is well chosen (after discovery/categorization, before job generation) and the reuse of `TestResults.not_run()` matching the render-only pattern is clean. A few things I noticed:

1. There's an exit code bug when `--dry-run` is used on a PyATS-only repo with no Robot templates. PyATS returns `TestResults.not_run("dry-run mode")` which has total=0, and since there's no Robot to contribute passed tests, `stats.is_empty` becomes True, which hits the `elif stats.is_empty: raise typer.Exit(1)` path in `main.py`. So the user gets exit code 1 and "No tests were executed" even though they explicitly asked not to execute tests. The render-only mode already handles this with an early return at `main.py:368` — dry-run probably needs a similar guard. The E2E test doesn't catch this because the mixed scenario has Robot templates that contribute 2 passed tests, keeping `stats.total > 0`.

2. The 6 unit tests in `test_orchestrator_dry_run.py` each repeat the same setup — env vars, directory creation, merged file, orchestrator instantiation. That same pattern actually appears about 18 times across the three orchestrator test files. A small `conftest.py` in `tests/pyats_core/` with a couple of fixtures (one for the directory structure, one for the credentials) would cut a lot of that boilerplate and make it easier to maintain if the orchestrator's constructor ever changes.

3. `test_orchestrator_accepts_dry_run_parameter` and `test_orchestrator_dry_run_defaults_to_false` only verify that Python's constructor assignment and default parameters work — the implementation is just `self.dry_run = dry_run`. These would pass even if all dry-run behavior was completely broken. The actual behavior is already well covered by the other 4 unit tests and the E2E suite, so these two aren't adding much safety.

4. The docstring in `combined_orchestrator.py` around line 84 still says `dry_run: Dry run mode (Robot only)` — since this PR extends dry-run to PyATS, that description is now stale.

5. `test_pyats_results_not_created` has an `if pyats_dir.exists():` guard, which means if the directory doesn't exist at all, the test passes without asserting anything. That could silently pass if something upstream is broken in a way that creates no output. Making the assertion unconditional (or at least explicitly handling both cases) would make the test more reliable.

6. Small one — the new code uses `"=" * 70` for separators while `summary_printer.py` uses `"=" * 80`. Not a big deal, just a minor visual inconsistency across the output.

What do you think?

_P.S. — This comment was drafted using voice-to-text via Claude Code. If the tone comes across as overly direct or terse, please know that's just how it tends to phrase things. No offense or criticism is intended — this is purely an objective technical review of the PR. Thanks for understanding! 🙂_

oboehmer:
Thanks for the thorough review @aitestino! All great catches. Here's how I've addressed each point:

**1. Exit code bug on PyATS-only dry-run** ✅
Fixed in `68075bb`. Added an early return for dry-run mode in `main.py` (similar to render-only), so it exits 0 before hitting the `stats.is_empty` check. Also added a new `DRY_RUN_PYATS_ONLY_SCENARIO` E2E test that would have caught this bug — it exercises the exact code path you identified.

**2. Test fixture consolidation** ✅
Created `tests/pyats_core/conftest.py` with shared fixtures (`clean_controller_env`, `aci_controller_env`, `sdwan_controller_env`, `cc_controller_env`, `pyats_test_dirs`) in `b61211e`. This eliminates the repeated setup across the orchestrator test files. Note: these fixtures intentionally duplicate some patterns from `tests/unit/conftest.py` — issue #541 will consolidate them when `tests/pyats_core/` merges into `tests/unit/`.

**3. Trivial constructor tests** ✅
Removed both `test_orchestrator_accepts_dry_run_parameter` and `test_orchestrator_dry_run_defaults_to_false` in `b61211e`. As you noted, the actual behavior is well covered by the other unit tests and E2E suite.

**4. Stale docstring** ✅
Fixed in `30eb0bb` — updated to `dry_run: Dry run mode (skips actual test execution)`.

**5. Unconditional assertion** ✅
Fixed in `aa03752`. The test now always asserts that the API/D2D result files don't exist in dry-run mode, rather than conditionally checking.

**6. Separator width**
Keeping `"=" * 70` since PR #566 removes `summary_printer` entirely, and we've aligned on this width for the dry-run output.

**Additional fix:** While testing, discovered that `pyats_results/` directory was being created even in dry-run mode (the directory tests were failing). Fixed in `68f961f` by moving the `mkdir` to after the dry-run check.

Ready for another look when you have a chance!

oboehmer:
@aitestino , could you please review this again? I addressed all the changes..

PR #563, opened 2026-02-21
feat: add standalone mock server management scripts

### LINKED ISSUE #562: test: support mock flask server outside pytest
The mock flask server introduced in #470 can come in handy for offline testing, so I created start/stop scripts to facilitate this use case..

oboehmer:
Add standalone mock server management:
- start_server.py: daemon mode with PID/log management in /tmp
- stop_server.py: graceful shutdown with fallback to SIGKILL
- status_server.py: health check with HTTP connectivity test
- Fix mock_server.py port handling to use self.port for standalone mode

Add mock API endpoints for scale testing:
- Update mock_api_config.yaml with 10 new vedges endpoints (vedges1-vedges10)
- Enables testing with multiple parallel test instances

Purpose: Enable standalone mock server usage outside pytest fixtures, for example performance profiling / scale testing.

Fixes #562


aitestino (changes requested):
Hey @oboehmer, nice addition — having standalone server management scripts will make manual E2E testing much smoother. A few things I noticed:

1. **240 lines of copy-pasted YAML (vedges1-vedges10)** — the 10 vedges endpoints in `mock_api_config.yaml` are identical except for the trailing number in the path. That's 240 lines of pure duplication. Could we use a single endpoint with regex/wildcard matching on the path (e.g., `/dataservice/system/device/vedges`), or use YAML anchors? If we need distinct paths for a specific test scenario, a comment explaining why would help.

2. **Duplicated constants across all 3 scripts** — `PID_FILE`, `DEFAULT_STANDALONE_PORT`, and the "read PID → validate → check process" logic are repeated in `start_server.py`, `stop_server.py`, and `status_server.py`. One option: consolidate into a single `mock_server_ctl.py` with argparse subcommands (`start`, `stop`, `status`). That eliminates the duplication entirely and gives you one entry point. Alternatively, extract shared constants/helpers to a small shared module.

3. **`status_server.py` hardcodes port 5555 for the health check** — if someone starts the server with `--port 8080`, the status script will still check `127.0.0.1:5555` and report "NOT RESPONDING." Consider writing the actual port to the PID file (e.g., `pid:port` format) or a small JSON state file so the status script knows where to look.

4. **Signal handler doesn't call `server.stop()`** — the `SIGTERM` handler in `start_server.py` cleans up the PID file and calls `sys.exit(0)`, but never calls `server.stop()`. The `KeyboardInterrupt` handler does call it though, so there's an inconsistency. The Flask thread may linger briefly on SIGTERM.

5. **Magic signal numbers** — `os.kill(pid, 15)` and `os.kill(pid, 9)` in `stop_server.py` would read better as `signal.SIGTERM` and `signal.SIGKILL`. Minor, but `signal` is already imported in `start_server.py` so the pattern is there.

What do you think?

_P.S. — This comment was drafted using voice-to-text via Claude Code. If the tone comes across as overly direct or terse, please know that's just how it tends to phrase things. No offense or criticism is intended — this is purely an objective technical review of the PR. Thanks for understanding! 🙂_

aitestino (changes requested):
Hey @oboehmer, nice addition — having standalone server management scripts will make manual E2E testing much smoother. A few things I noticed:

1. **240 lines of copy-pasted YAML (vedges1-vedges10)** — the 10 vedges endpoints in `mock_api_config.yaml` are identical except for the trailing number in the path. That's 240 lines of pure duplication. Could we use a single endpoint with regex/wildcard matching on the path (e.g., `/dataservice/system/device/vedges`), or use YAML anchors? If we need distinct paths for a specific test scenario, a comment explaining why would help.

2. **Duplicated constants across all 3 scripts** — `PID_FILE`, `DEFAULT_STANDALONE_PORT`, and the "read PID → validate → check process" logic are repeated in `start_server.py`, `stop_server.py`, and `status_server.py`. One option: consolidate into a single `mock_server_ctl.py` with argparse subcommands (`start`, `stop`, `status`). That eliminates the duplication entirely and gives you one entry point. Alternatively, extract shared constants/helpers to a small shared module.

3. **`status_server.py` hardcodes port 5555 for the health check** — if someone starts the server with `--port 8080`, the status script will still check `127.0.0.1:5555` and report "NOT RESPONDING." Consider writing the actual port to the PID file (e.g., `pid:port` format) or a small JSON state file so the status script knows where to look.

4. **Signal handler doesn't call `server.stop()`** — the `SIGTERM` handler in `start_server.py` cleans up the PID file and calls `sys.exit(0)`, but never calls `server.stop()`. The `KeyboardInterrupt` handler does call it though, so there's an inconsistency. The Flask thread may linger briefly on SIGTERM.

5. **Magic signal numbers** — `os.kill(pid, 15)` and `os.kill(pid, 9)` in `stop_server.py` would read better as `signal.SIGTERM` and `signal.SIGKILL`. Minor, but `signal` is already imported in `start_server.py` so the pattern is there.

What do you think?

_P.S. — This comment was drafted using voice-to-text via Claude Code. If the tone comes across as overly direct or terse, please know that's just how it tends to phrase things. No offense or criticism is intended — this is purely an objective technical review of the PR. Thanks for understanding! 🙂_

oboehmer (commented):
Thanks for the thorough review! I've addressed all 5 points:

1. **240 lines of YAML duplication (vedges1-vedges10)** → Consolidated into a single regex-matched endpoint using `path_pattern: "^/dataservice/system/device/vedges[0-9]+$"` with `match_type: "regex"`. Reduced ~240 lines to ~30 lines.

2. **Duplicated constants across 3 scripts** → Consolidated `start_server.py`, `stop_server.py`, and `status_server.py` into a single `mock_server_ctl.py` with argparse subcommands (`start`, `stop`, `status`). One entry point, no duplication.

3. **status_server.py hardcodes port 5555** → Solved by the consolidation above. The state file now stores both PID and port as JSON (`{"pid": 12345, "port": 8080}`), so `status` always checks the correct port.

4. **SIGTERM handler doesn't call server.stop()** → Fixed. The signal handler now calls `server.stop()` before cleanup, consistent with the KeyboardInterrupt handler.

5. **Magic signal numbers (15, 9)** → Replaced with `signal.SIGTERM` and `signal.SIGKILL` constants for readability.

Also added a `README.md` in the mocks directory with usage instructions.

@aitestino Ready for re-review when you have a moment!

PR #657, opened 2026-03-17
feat(pyats): add tag-based filtering for pyATS tests

### LINKED ISSUE #436: [feat] support filtering pyats tests using tags
## Goal

Enable include/exclude filtering for pyATS tests in nac-test, similar to how Robot tests can be filtered with `--include`/`--exclude` tags (see [Robot User Guide](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#by-tag-names))

## TL;DR

- pyATS supports **`groups`** - a simple class attribute on test classes (list of strings)
- We can think about generating this attribute as part of QS generation, but also enable users to easily modify/customize this themselves
- Recommended: Use `groups` for filtering 
- Can benefit our own testing as well (to only run specific test cases), so happy to implement this myself

## Implementation Goals:

- Align to pyats filtering markers (even if our test scripts will never be executed outside `nac-test`)
- Efficient implementation, perform filtering in nac-test during TestDiscovery to avoid subprocess overhead

## How pyATS Groups Work

Test classes can declare group membership:

```python
class Verify_Foo(IOSXETestCase):
    groups = ['health', 'bgp', 'longrunning']
```

Then filter at runtime: include only `health` tests, or exclude long running tests.

## Proposed Approach

| Step | Description |
|------|-------------|
| 1. TestDiscovery parses groups during pyats discovery (in PyatsOrchstrator) | Extract `groups` attribute from test classes via AST |
| 2. Pre-filter test files  | Skip files where no test classes match the filter expression before d2d/api classification |
| 3. Pass tests to pyATS runtime | Unchanged logic, need to gracefully handle no tests matching filter expression |

**Benefits:**
- ✅ Native pyATS mechanism (no custom filter markers)
- ✅ Matches nac-test's existing test generation
- ✅ Consistent UX with Robot's tag filtering

**Trade-offs:**
- ⚠️ Groups must be statically declared (no runtime assignment)
- ⚠️ Adds complexity to discovery phase

## Alternative pyATS Filtering Options

pyATS also supports UID-based filtering:

| Mechanism | Description | Use Case |
|-----------|-------------|----------|
| `uids` | Include test sections by UID pattern | Filter specific testcases by class name |
| `exclude_uids` | Exclude test sections by UID pattern | Skip specific testcases |

**UIDs** default to the class name (e.g., `MyTestcase`) and can be filtered with regex patterns or Logic expressions (`And`, `Or`, `Not`).

**Why `groups` is preferred for nac-test:**
- Groups are semantic ("sanity", "slow") vs UIDs which are structural (class names)
- Groups allow multiple categories per test; UIDs are single identifiers
- Groups align better with Robot's tag-based filtering model

## Open Questions

| Question | Options |
|----------|---------|
| CLI interface | New flags (`--include-groups`) vs extend existing (`--include`) |
| Filtering Logic | Same logic/semantic for Robot + pyATS, or separate? |

## Next Steps

1. Decide on CLI interface approach and filtering logic
2. Implement group extraction in discovery

## Open Items (not in scope of this ticket)

- How to report tags?

oboehmer:
## Description

This PR adds tag-based filtering support for PyATS tests using the `--include` and `--exclude` CLI options, bringing PyATS test selection in line with Robot Framework's tag filtering capabilities. PyATS tests are filtered based on their `groups` class attribute using Robot Framework's `TagPatterns` API for consistent pattern matching across both frameworks.

## Closes

- Fixes #436

## Related Issue(s)

- N/A

## Type of Change

- [x] Documentation update

## Test Framework Affected

- [x] PyATS
- [x] Robot Framework
- [ ] Both

- [ ] ACI (APIC)
- [ ] Meraki (Cloud-managed)
- [ ] Hyperfabric
- [x] All architectures

## Platform Tested

- [x] macOS (version tested: )

### Core Tag Filtering Implementation

- **New `TagMatcher` class** (`tag_matcher.py`): Wraps Robot Framework's `TagPatterns` API to provide consistent tag matching behavior for both Robot and PyATS tests
  - Supports all Robot Framework tag pattern syntax: simple tags, wildcards (`bgp*`), boolean operators (`healthANDbgp`, `bgpORospf`, `bgpNOTnrfu`)
  - Case-insensitive matching with underscore normalization
  - Include/exclude patterns are private; no public instance state

- **Refactored test discovery architecture** (`test_discovery.py`, `test_type_resolver.py`):
  - Renamed `TestTypeResolver` to `TestMetadataResolver` to reflect expanded responsibilities
  - AST-based extraction of `groups` attribute from PyATS test classes without importing
  - Pre-filtering during discovery phase eliminates subprocess overhead
  - Combined discovery and categorization into single pass returning `PyatsDiscoveryResult`
  - `TestMetadataResolver` methods converted to `@staticmethod`; no instantiation required. Module-level logger replaces `self.logger`, consistent with the rest of the codebase

- **Updated dataclasses** (`types.py`):
  - `TestFileMetadata`: Structured representation of test file metadata (path, type, groups)
  - `PyatsDiscoveryResult`: Discovery results container with computed properties (`total_count`, `all_tests`, `api_paths`, `d2d_paths`). The former `test_type_by_path` derived dict and `get_test_type()` method have been removed -- type information is now carried forward through the execution pipeline rather than reconstructed after the fact (see Execution Pipeline below)

- **Execution pipeline** (`orchestrator.py`, `output_processor.py`):
  - `OutputProcessor` receives a `test_type_by_path` map at construction time and stamps `test_type` directly into each `test_info` entry at `task_start` time
  - The post-execution split loop reads `test_info["test_type"]` directly -- no external path lookup needed
  - This removes the design smell where type information was known at discovery but dropped and re-derived after execution
  - A comment in the orchestrator documents the deeper future fix: separate `OutputProcessor`/`test_status` instances per execution type

- **Orchestrator integration** (`orchestrator.py`):
  - Improved error messaging: "No pyATS tests matching tag filter (exclude: 'bgp OR ospf')" when all tests filtered out
  - Distinguishes between "no test files" vs "no tests matching filter"
  - Discovery print restored to show breakdown: "Discovered N PyATS test files (M api, K d2d)"

- **Documentation** (`README.md`, `PRD_AND_ARCHITECTURE.md`):
  - Updated `--include`/`--exclude` CLI option descriptions
  - Added PyATS `groups` attribute usage examples
  - Tag pattern syntax reference

#### New Tests for Tag Filtering

- **E2E tests** (`tests/e2e/`):
  - 4 new scenarios covering include/exclude combinations
  - Fixtures with both Robot and PyATS tests tagged for filtering validation
  - Tests verify correct exit codes and test counts for various filter patterns
  - Clarified scenario descriptions: exit 252 ("no tests found") is only raised when the combined total across all frameworks is zero

- **Unit tests** (`tests/unit/pyats_core/discovery/`):
  - `test_tag_matcher.py`: Tag pattern matching, boolean logic, wildcards, edge cases
  - `test_metadata_resolver.py`: AST-based test type detection, directory fallback, error handling
  - `test_groups_extraction.py`: Groups attribute extraction for tag filtering
  - Updated orchestrator tests to use new `PyatsDiscoveryResult` structure

- **New unit tests for previously untested code**:
  - `tests/unit/pyats_core/common/test_types.py`: `PyatsDiscoveryResult` computed properties (`total_count`, `all_tests`, `api_paths`, `d2d_paths`)
  - `tests/unit/pyats_core/execution/test_output_processor.py`: `test_type` stamping for known api/d2d paths, unknown paths, absent map, and missing `test_file` in event

#### Test Suite Refactoring (Independent of Feature)

**Consolidation: Integration to Unit Tests**
- Moved 42 non-integration tests from `test_discovery_integration.py` to new `tests/unit/pyats_core/discovery/test_test_discovery.py`
- Retained only `TestDiscoveryPerformance` as the true integration test (validates timing with 100+ files)
- Result: Integration tests reduced from 43 to 1

**Parametrization**
- Consolidated 39 individual test methods in `test_tag_matcher.py` into 15 parametrized tests (86 total cases)
- Applied pytest parametrization per project testing guidelines

**Code Quality**
- Added proper type annotations for all test parameters
- Fixed mypy errors for generic types (`list` to `list[str]`)
- Added `MockerFixture` type hints for pytest-mock fixtures
- Flattened nested `with patch.object(...):` blocks using Python 3.10+ parenthesised multi-target `with` statements

> **Note**: This refactoring improves test maintainability and follows project testing guidelines but is not strictly required for the tag filtering feature.

## Testing Done

  - `tests/unit/pyats_core/discovery/`: tag matcher, metadata resolver, groups extraction
  - `tests/unit/pyats_core/common/test_types.py`: `PyatsDiscoveryResult` computed properties
  - `tests/unit/pyats_core/execution/test_output_processor.py`: `test_type` stamping behaviour
  - Updated orchestrator tests to use new `PyatsDiscoveryResult` structure
- [x] E2E tests added
  - 4 tag filtering scenarios validating include/exclude combinations

### Test Commands Used

```bash
# Run all unit and integration tests
uv run pytest -n auto --dist loadscope

# Test tag filtering - include specific tag
nac-test -d ./tests/e2e/fixtures/tag_filtering -t ./tests/e2e/fixtures/tag_filtering/templates -o ./output --include bgp

# Test tag filtering - exclude specific tag
nac-test -d ./tests/e2e/fixtures/tag_filtering -t ./tests/e2e/fixtures/tag_filtering/templates -o ./output --exclude ospf

# Test tag filtering - boolean pattern
nac-test -d ./data -t ./tests -o ./output --include "healthORbgp" --exclude nrfu

# Test no matches (exits 0 when Robot tests still run)
nac-test -d ./tests/e2e/fixtures/tag_filtering -t ./tests/e2e/fixtures/tag_filtering/templates -o ./output --exclude "bgpORospf"
```

## Checklist

- [x] Documentation updated (if applicable)
- [x] CHANGELOG.md updated (if applicable)

N/A

### Implementation Approach

The tag filtering implementation leverages Robot Framework's `TagPatterns` API to ensure consistent behavior between Robot and PyATS test selection. Key design decisions:

1. **PyATS `groups` attribute**: Uses the native pyATS mechanism (simple list of strings on test classes) for tagging
   ```python
   class VerifyBgpNeighbors(SDWANTestBase):
       groups = ["bgp", "routing"]  # Matches --include bgp or --include routing
   ```

2. **Discovery-phase filtering**: Tests are filtered during the discovery phase via AST parsing, avoiding subprocess overhead and enabling early validation

3. **Single-pass architecture**: Discovery and categorization combined into one pass that returns a `PyatsDiscoveryResult` with pre-computed properties for efficient access

4. **Type information carried forward**: `test_type` is stamped into each result entry at the moment it is first observed (task_start), rather than being re-derived from path lookups after execution completes

5. **Consistent pattern formatting**: Uses Robot Framework's `TagPatterns` string representation for user-facing messages (`'bgpORospf'` to `'bgp OR ospf'`)

### Breaking Changes

None. The changes are backward-compatible:
- PyATS tests without `groups` attribute remain unaffected and run as before
- Existing `--include`/`--exclude` behavior for Robot Framework unchanged
- All existing test repositories continue to work without modification

### Migration Notes

- PyATS tests can optionally add a `groups` class attribute to enable tag-based filtering
- The `groups` attribute should be a list of strings (native pyATS format)
- Tests without `groups` are treated as having no tags and are included by default (unless an `--include` pattern is specified)


aitestino (changes requested):
Hey @oboehmer, thank you for the PR -- this is a well-designed feature that fills a real gap in the pyATS side of the tool. Reusing Robot Framework's `TagPatterns` for consistent pattern matching across both frameworks is the right call, and the single-pass discovery refactoring is a genuine simplification over the old two-phase approach. The `PyatsDiscoveryResult` dataclass replacing the opaque `tuple[list[Path], list[tuple[Path, str]]]` return type is a nice readability win, and I appreciate that the issue's open questions (extend existing `--include`/`--exclude` vs new flags, same logic vs separate) are resolved cleanly -- extending the existing flags with shared TagPatterns semantics is exactly the right direction.

I did a couple of reviews on this, and I think we need some small changes before merge. The core design is sound -- these are all polish items.

1. **`DEFAULT_TEST_TYPE` missing `Final` annotation** -- The constant was moved from `test_type_resolver.py` (where it was `Final[str]`) to `common/types.py` (where it's just `str`). The `Final` annotation should be preserved so mypy catches accidental reassignment. One-liner fix.

2. **`helpers.py` should be `conftest.py`** -- `tests/unit/pyats_core/discovery/helpers.py` introduces a new convention for shared test utilities. Looking at the existing codebase, we consistently use `conftest.py` for this purpose (there are 5 conftest files across the test tree already). Renaming it would keep the pattern consistent.

3. **Duplicate XML-parsing in E2E tag filter tests** -- `test_only_bgp_robot_test_in_output_xml` and `test_ospf_robot_test_excluded_from_output_xml` share identical `ET.parse` -> `findall(".//test")` -> extract names logic. A small helper like `_get_robot_test_names(results)` would clean this up -- maybe 5 lines.

4. **`slots=True` inconsistency** -- `TestFileMetadata` uses `@dataclass(slots=True)` but `PyatsDiscoveryResult` uses bare `@dataclass`. Either apply consistently or document the intentional difference.

5. **`test_type` could benefit from `Literal["api", "d2d"]`** -- `TestFileMetadata.test_type` is typed as `str` but only accepts two values. A `Literal["api", "d2d"]` would close the gap for mypy and prevent silent misuse.

6. **`pyproject.toml` filterwarnings may be unnecessary** -- The `norecursedirs = ["nac_test"]` addition already prevents pytest collection of `Test*` source classes. The `filterwarnings` regex will need manual updates as new classes are added -- consider whether `norecursedirs` alone is sufficient.

7. **Missing E2E coverage for wildcard/boolean patterns** -- The README documents `bgp*`, `bgpANDospf`, `bgpORospf` as supported but no E2E test exercises these patterns. The unit-level TagMatcher coverage is strong, but a single E2E scenario with a wildcard or boolean include would close the loop.

8. **`TagMatcher.__str__` iterates `TagPatterns` directly** -- The `", ".join(f"'{p}'" for p in self._include_patterns)` relies on implicit iteration behavior of a third-party object. Could use `self._include_list` instead, which is already stored and unambiguously a `list[str]`.

9. **Two tautological tests in `test_types.py`** -- `test_api_paths_excludes_d2d` and `test_d2d_paths_excludes_api` assert that data not put into a list isn't in the list. They test the fixture setup, not the property logic. Consider removing them.

10. **Redundant assertions in `test_resolve_returns_metadata_with_groups`** -- The `isinstance(result, TestFileMetadata)` and `hasattr(result, "groups")` checks are already enforced by mypy. Only the `assert result.groups == ["health", "bgp"]` line is load-bearing.

11. **Repeated `pytest.skip` guards** -- Four identical `pytest.skip("No tests expected - combined summary not generated")` guards in `test_e2e_scenarios.py` could be consolidated into the base class or a shared property check.

I'd consider items 1-3 as things to address before merge -- they're all quick fixes. Items 4-11 are lower priority and could be follow-ups if you'd rather not hold this up further.

What do you think?

P.S. -- Once this lands, could you also open a PR against the Quicksilver test generation templates to update them so that any newly produced test cases include the `groups` attribute following this design pattern? That way new tests get tagging support out of the box.

*P.P.S. -- This comment was drafted using voice-to-text via Claude Code. If the tone comes across as overly direct or terse, please know that's just how it tends to phrase things. No offense or criticism is intended -- this is purely an objective technical review of the PR. Thanks for understanding! 🙂*

oboehmer:
Thanks for the thorough review! Addressed below:

**1. `Final` annotation** — I'd like to handle this holistically rather than adding `Final` to just this constant — `Final` is currently only used in `test_type_resolver.py`, while `core/constants.py` and other modules don't use it. Adding it here would widen rather than narrow the inconsistency. Opened #760 to discuss the right approach across the codebase.

**2. `helpers.py` → `conftest.py`** — Done. 8b346ec

**3. Duplicate XML-parsing** — Extracted into a module-level `_get_robot_test_names()` helper in the same file, using the existing top-level `ROBOT_RESULTS_DIRNAME` import. 940048f

**4. `slots=True` inconsistency** — This is intentional: `TestFileMetadata` can have thousands of instances during discovery of large test repositories, while `PyatsDiscoveryResult` is a singleton per run. Added a docstring explaining the rationale. cd5573c

**5. `Literal["api", "d2d"]`** — Done. Introduced a `TestType` alias and applied it through the chain (`TestFileMetadata`, `DEFAULT_TEST_TYPE`, `BASE_CLASS_MAPPING`, `VALID_TEST_TYPES`, `_detect_via_directory`). Also updated `OutputProcessor.test_type_by_path` to use `Mapping[str, str]` for covariance. a3a2d5a, 636792d

**6. `filterwarnings`** — `norecursedirs` prevents pytest from *recursing* into `nac_test/` during collection, but it doesn't prevent warnings when test modules *import* source classes matching `Test*`. The `filterwarnings` regex is needed specifically for those imported symbols (e.g. `TestFileMetadata`, `TestMetadataResolver`). Both are complementary.

**7. Wildcard/boolean E2E coverage** — Rather than adding a new scenario (each adds 10+ secs), changed the existing exclude fixture from `--exclude ospf` to `--exclude osp*`. Same test expectations, exercises wildcard matching at the E2E level. 4fafac8

**8. `TagMatcher.__str__` iterating `TagPatterns`** — The iteration is deliberate: `__str__` is the user-facing form and iterating `TagPatterns` yields Robot's canonical formatted representation (e.g. `"bgpORospf"` → `"bgp OR ospf"`). Switching to `_include_list` would lose this formatting and degrade the user-facing messages. `__repr__` already provides the raw input form. Added a docstring comment making this intent explicit. a9ef3be

**9. Tautological tests** — Removed. a8c6135

**10. Redundant assertions** — Removed `isinstance` and `hasattr` checks; only the value assertion is load-bearing. 6b2833e

**11. Repeated `pytest.skip` guards** — These are intentional — each skip is a per-test-method decision based on scenario capabilities, and this pattern is established across virtually all base class test methods (~50 uses). Consolidating into the base class would either require introspective property checks or make the skip logic less transparent per test.

oboehmer:
Merging this now as all comments have been addressed and you're on holiday. happy to address changes in follow-up PRs.

Issue #677, opened 2026-03-20
Remove merged data model artifact after test execution to prevent potential credential exposure


oboehmer:
## Problem Statement

During test execution, `nac-test` creates a `merged_data_model_test_variables.yaml` artifact in the output directory. This file contains the fully-rendered data model used by pyATS test cases.

**Important context on what this file does NOT contain:**
- Controller API credentials (ACI_PASSWORD, SDWAN_PASSWORD, etc.) — these are read from environment variables at runtime and never written to disk
- Device SSH credentials — these flow through the testbed generator, not the merged data model

**What this file MAY contain:**
- Values resolved from `!env` and `!vault` tags in the user's data model
- If users store sensitive data (e.g., SNMP community strings, application credentials, API keys for non-controller systems) in their data model using these tags, the resolved plaintext values end up in this file

While the primary authentication credentials are already handled securely via environment variables, the merged data model artifact could inadvertently contain other sensitive information depending on the user's data model content. This poses a concern if:
- The output directory is archived by CI/CD pipelines
- Build artifacts are retained or shared
- The file is left behind after execution

## Proposed Solution

**Phase 1: Automatic cleanup (recommended)**

1. **Delete the file immediately after pyATS execution completes**
   - Remove in `main.py` after the pyATS subprocess returns
   - Register an `atexit` handler as a safety net for interrupted executions
   - Handle `SIGTERM`/`SIGINT` for graceful cleanup on forced termination

2. **Restrict file permissions while it exists**
   - Create with `0o600` (owner read/write only)
   - Reduces exposure window during execution

**Phase 2: In-memory serving via broker (optional future enhancement)**

For environments with stricter security requirements, extend the existing `ConnectionBroker` IPC to serve the data model in-memory:
- pyATS tests request data model via Unix socket instead of file read
- No file ever written to disk
- Leverages existing broker infrastructure

## Alternatives Considered

| Approach | Pros | Cons |
|----------|------|------|
| Delete after execution | Simple, no architecture change | Brief exposure window |
| Broker-based serving | Zero disk exposure | More complex, overkill for most cases |
| Documentation only | Zero effort | Not acceptable — doesn't address the risk |
| Encrypt file at rest | Data protected on disk | Key management complexity, still on disk |

## Implementation Notes

- The file is written in `DataMerger.write_merged_data_model()` called from `main.py:377`
[...]

Issue #538, opened 2026-02-16
refactor: consolidate PYATS_DEBUG / KEEP_HTML_REPORT_DATA environment variable checks


aitestino:
## Summary

The debug file retention logic — checking `PYATS_DEBUG` and `KEEP_HTML_REPORT_DATA` environment variables — is duplicated across multiple reporting modules with slightly different implementations. This should be consolidated into a shared utility.

## Current Duplication

1. **`nac_test/pyats_core/reporting/generator.py`** — checks both env vars, logs which one is set
2. **`nac_test/pyats_core/reporting/multi_archive_generator.py`** — checks both env vars, logs differently
3. **`nac_test/pyats_core/orchestrator.py`** — references the same env vars

Each location has slightly different logging messages and conditional structures for the same decision.

## Proposed Solution

Extract to a shared utility:

```python
# nac_test/utils/debug.py
import logging
import os

logger = logging.getLogger(__name__)

def should_keep_debug_files() -> bool:
    """Check if debug files should be kept based on environment variables.

    Returns:
        True if PYATS_DEBUG or KEEP_HTML_REPORT_DATA is set.
    """
    return bool(
        os.environ.get("PYATS_DEBUG") or os.environ.get("KEEP_HTML_REPORT_DATA")
    )

def log_debug_file_decision(keeping: bool) -> None:
    """Log the debug file retention decision with the reason."""
    if keeping:
        reason = "PYATS_DEBUG" if os.environ.get("PYATS_DEBUG") else "KEEP_HTML_REPORT_DATA"
        logger.info(f"Keeping debug files ({reason} is set)")
    else:
        logger.info("Cleaning up debug files")
```

## Acceptance Criteria

- [ ] Shared utility created in `nac_test/utils/debug.py`
- [ ] All 3 locations refactored to use shared utility
- [ ] Consistent logging messages across all call sites
- [ ] Existing tests pass without modification

PR #697, opened 2026-03-22
fix(pyats): refactor pyats config file creation (incl disable EnvironmentDebugPlugin)

### LINKED ISSUE #570: Tech Debt: Consolidate PyATS config file creation in SubprocessRunner
## Context

In `nac_test/pyats_core/execution/subprocess_runner.py`, the `SubprocessRunner` class creates temporary config files (plugin config and pyats config) in both `execute_job()` and `execute_job_with_testbed()`. This code is duplicated and the config content is static.

Additionally, there is an unused `plugin_config_path` instance attribute with unclear intent:
1. The orchestrator generates a **reporter config** via `_generate_plugin_config()` (configures QueueHandler/FileHandler for async reporting)
2. This is stored in `subprocess_runner.plugin_config_path`
3. However, `SubprocessRunner` ignores this and creates its own **plugin config** (enables ProgressReporterPlugin)
4. Both configs use the same `--configuration` CLI flag, so only one can be used

## Suggested Improvements

1. **Extract config file creation into helper methods**:
   - `_create_plugin_config_file()` - static method
   - `_create_pyats_config_file()` - static method

2. **Create config files once in `__init__()`** rather than per-job:
   - Since the config content is static, files can be created once and reused
   - This reduces I/O overhead when running multiple jobs

3. **Clarify the relationship between reporter config and plugin config**:
   - Determine if the orchestrator reporter config should be merged with the plugin config
   - If both are needed, merge them into a single config file
   - If the reporter config is obsolete, remove `_generate_plugin_config()` and the `plugin_config_path` parameter

## Related

This was identified during PR #557 review but deemed out of scope for that change.

## Files

- `nac_test/pyats_core/execution/subprocess_runner.py`
- `nac_test/pyats_core/orchestrator.py` (generates reporter config via `_build_reporter_config()`)

### LINKED ISSUE #689: Security: Disable pyATS EnvironmentDebugPlugin to prevent credential exposure in env.txt
## Summary

PyATS's `EnvironmentDebugPlugin` (enabled by default) creates an `env.txt` file in the results archive that dumps **all environment variables in cleartext**, exposing sensitive credentials.

## Problem

When running PyATS tests via nac-test, the generated `pyats_results` archives contain an `env.txt` file with all environment variables, including:
- Controller credentials (`ACI_PASSWORD`, `SDWAN_PASSWORD`, `CC_PASSWORD`)
- Device credentials (`IOSXE_PASSWORD`)
- API tokens and secrets

This is particularly dangerous in CI/CD environments where test artifacts may be stored, shared, or uploaded.

## Solution

Disable the `EnvironmentDebugPlugin` in the easypy plugin configuration that nac-test generates for PyATS subprocess execution.

**File:** `nac_test/pyats_core/execution/subprocess_runner.py`

The plugin configuration would include:
```yaml
plugins:
    ProgressReporterPlugin:
        enabled: True
        module: nac_test.pyats_core.progress.plugin
        order: 1.0
    EnvironmentDebugPlugin:
        enabled: False
```

## Related

- Upstream pyATS issue: https://wwwin-github.cisco.com/pyATS/pyats/issues/895

oboehmer:
## Description

Security fix: PyATS EnvironmentDebugPlugin was writing environment variables (including passwords) to `env.txt` files within archive artifacts. This change disables the plugin and refactors config file management for cleaner architecture.
Also refactor pyats config file creation and remove dead code

## Closes

- Fixes #689
- Fixes #570

## Related Issue(s)

- #677 (planned robust cleanup mechanism; `__del__` in this PR is a stopgap)

## Type of Change

- [ ] Documentation update

## Test Framework Affected

- [x] PyATS
- [ ] Robot Framework
- [ ] Both

- [ ] ACI (APIC)
- [ ] Meraki (Cloud-managed)
- [ ] Hyperfabric
- [ ] All architectures
- [x] N/A (architecture-agnostic)

## Platform Tested

- [x] macOS (version tested: macOS with Python 3.12)

### Security (#689)
- Add `EnvironmentDebugPlugin: enabled: False` to plugin configuration
- Add E2E regression test scanning ALL output artifacts for credential sentinels
- Test covers both zip contents and all non-zip files

### Architecture Refactoring (#570)
- Move config file creation from `execute_job()` into `SubprocessRunner.__init__()`
- Add explicit `cleanup()` method to SubprocessRunner for config file removal
- Add `__del__` for opportunistic cleanup on unexpected exits (best-effort; see #677)
- Extract `_cleanup_subprocess_runner()` helper in orchestrator to DRY up two identical call sites
- Orchestrator calls `cleanup()` conditionally based on keep_artifacts policy
- Config file attributes only set after successful file creation (safer error handling)
- Extract config filenames as constants in `constants.py`
- Write config files to output_dir with well-known names instead of temp files

### Dead Code Removal
- Remove unused `_build_reporter_config()` and `_generate_plugin_config()` from orchestrator
- Remove unnecessary `tempfile.TemporaryDirectory` wrapper in `_run_tests_async`
- Remove unused `plugin_config_path` parameter

## Testing Done

- [x] Unit tests added/updated (26 subprocess_runner unit tests)
- [x] Integration tests performed (16 subprocess_runner integration tests)

### Test Commands Used

```bash
# Unit tests (26 subprocess_runner tests, 219 pyats_core tests total)
pytest tests/unit/pyats_core/execution/test_subprocess_runner.py -v
pytest tests/unit/pyats_core/ tests/pyats_core/ -v

# E2E credential exposure tests
pytest tests/e2e/test_e2e_scenarios.py -v -k "credential" -n auto --dist loadscope

# Pre-commit hooks
pre-commit run --all-files
```

## Checklist

N/A

## Additional Notes

**Design decisions:**
- `SubprocessRunner` now owns both creation AND cleanup of its config files (cleaner SRP)
- Orchestrator retains policy control (decides when to call cleanup based on debug flags)
- `__del__` cleanup is a safe no-op if files are already gone or were never created
- Config file cleanup is best-effort: files are small and non-sensitive; a robust cleanup mechanism covering all exit paths will be added as part of #677


aitestino (changes requested):
Hey @oboehmer, thank you for the PR -- this is a well-motivated change that addresses a real security gap (EnvironmentDebugPlugin leaking credentials into archive artifacts) and takes the opportunity to clean up genuine architectural debt along the way. Moving config file ownership into SubprocessRunner so it owns both creation and cleanup is the right SRP boundary, the DRY win from eliminating those two copy-pasted config blocks in `execute_job` and `execute_job_with_testbed` is substantial (especially since they had already diverged -- one was missing the security fix), and the dead code removal of `_build_reporter_config` and `_generate_plugin_config` is overdue -- those methods were not only unused but broken (the `dump_to_stream` import references a module that doesn't exist). The E2E credential sentinel test is a strong regression guard, and the partial-write cleanup logic in `_create_config_files` is thoughtfully implemented.

I ran this through a pretty thorough review, and I think we need some adjustments before merge. The core design is sound and the security fix is correct (I traced through the PyATS source to confirm `recursive_update` properly merges the `enabled: False` override) -- these are all refinement items.

1. **Cleanup gap on "no archives found" early return** -- In `_generate_html_reports_async`, the early return when `archive_paths` is empty skips `_cleanup_subprocess_runner()` entirely. The `keep_artifacts` variable is computed just above that block, so the fix is one line: `self._cleanup_subprocess_runner(keep_artifacts)` before the return. Alternatively, the two existing `_cleanup_subprocess_runner` call sites on the success and failure branches could be collapsed into a single `try/finally` -- that would also prevent this class of bug if a third return path is ever added.

2. **`PLUGIN_CONFIG` and `PYATS_CONFIG` should live in `constants.py`** -- We try to centralize constants in `constants.py` -- the filename constants (`PYATS_PLUGIN_CONFIG_FILENAME`, `PYATS_CONFIG_FILENAME`) are correctly placed there, but the content constants ended up in `subprocess_runner.py`. Moving them to `constants.py` as `PYATS_PLUGIN_CONFIG_CONTENT: str` and `PYATS_CONFIG_CONTENT: str` (with type annotations, matching the other constants in that file) would keep everything in one place. As a bonus, anyone auditing the plugin configuration finds filenames and their contents together.

3. **New constants missing from `__all__` in `constants.py`** -- `PYATS_PLUGIN_CONFIG_FILENAME` and `PYATS_CONFIG_FILENAME` aren't added to the `__all__` list that starts around line 113. Quick one-liner fix.

4. **`Path.write_text` patches are too broad in both test files** -- `patch.object(Path, "write_text", ...)` intercepts every `Path.write_text` call in the process, not just the one in `_create_config_files`. For `test_orchestrator_config_error.py`, patching at `SubprocessRunner._create_config_files` with `side_effect=RuntimeError(...)` is more precise and decouples from the OSError-to-RuntimeError translation already covered in unit tests. For the unit tests, scoping to `"nac_test.pyats_core.execution.subprocess_runner.Path.write_text"` would be the right target.

5. **`test_del_calls_cleanup` tests the wrong contract** -- It calls `runner.__del__()` and asserts files are removed, but `test_cleanup_removes_config_files` already covers that exact behavior. The actual `__del__` contract -- the thing that distinguishes it from calling `cleanup()` directly -- is that it swallows exceptions silently. Replacing with a test that patches `cleanup()` to raise and verifies `__del__()` does not propagate would test the real behavior.

6. **`NAC_TEST_PYATS_KEEP_REPORT_DATA` env var string is scattered across 3 source files** -- `orchestrator.py`, `multi_archive_generator.py`, and `generator.py` all evaluate `os.environ.get("NAC_TEST_PYATS_KEEP_REPORT_DATA")` independently. The existing `DEBUG_MODE = get_bool_env("NAC_TEST_DEBUG")` pattern shows exactly how to consolidate this: `KEEP_REPORT_DATA: bool = get_bool_env("NAC_TEST_PYATS_KEEP_REPORT_DATA")` in `constants.py`. The PR's `keep_artifacts` local variable is a step in the right direction -- this would complete it. Pre-existing debt, but the PR adds a new evaluation path so now is a natural moment.

7. **`object.__new__(SubprocessRunner)` in two unit tests creates fragile coupling** -- `test_write_failure_leaves_attributes_none_and_cleanup_is_safe` and `test_partial_write_failure_cleans_up_first_file` bypass `__init__` to test `_create_config_files` in isolation. This couples to internal attribute names and creates partially-initialized objects that can't exist in production. For the partial-write test, the value is anchored in the filesystem assertion so it's tolerable. For the attribute-state test, consider either mocking the pyats executable check or folding the assertion into the existing `test_init_raises_runtime_error_on_write_failure`.

8. **Missing `_cleanup_subprocess_runner` policy tests** -- The helper has three behaviors worth covering: (a) no-op when `subprocess_runner` is None, (b) skips cleanup when `keep_artifacts=True`, (c) delegates to `cleanup()` when both conditions are met. Three quick tests, each about 5 lines.

9. **Test file placement deepens Issue #541 debt** -- `test_orchestrator_config_error.py` is placed in `tests/pyats_core/`, but that directory's own `conftest.py` documents the intent to merge `tests/pyats_core/` into `tests/unit/`. Placing new files there moves in the wrong direction. Worth considering `tests/unit/pyats_core/` instead, or at minimum acknowledging this as #541 scope.

10. **Security advisory: consider disabling `ArchiveConfigPlugin` in a follow-up** -- The `ArchiveConfigPlugin` (also in PyATS's `debug.py`) remains enabled and writes `pyats.configuration.yaml` and `easypy.configuration.yaml` to archives. For nac-test's current usage the risk is limited since controller credentials flow via environment variables, but for D2D tests with testbed files, device credentials could potentially be reflected. Adding `ArchiveConfigPlugin: enabled: False` to the plugin config would provide defense-in-depth. Not blocking -- the E2E sentinel test would catch it if credentials leak through this path.

I'd consider items 1-5 as things to address before merge -- they're all quick fixes (the cleanup gap is the most important one). Items 6-10 are lower priority and could be follow-ups if you'd rather not hold this up further.

What do you think?

oboehmer:
Thanks for the thorough review @aitestino! Here's how we addressed each item:

| # | Comment | Status | Detail |
|---|---------|--------|--------|
| **1** | Cleanup gap on early return | ✅ Resolved | Refactored to use `CleanupManager` (from #727) — cleanup is now automatic via atexit/signals, no manual call sites to miss |
| **2** | Content constants to `constants.py` | ⏭️ Deferred | Config file content in `constants.py` looks odd — these are multi-line YAML templates, not simple constants. Prefer to keep them co-located with their only consumer |
| **3** | Missing `__all__` entries | ✅ Fixed | Added 3 entries |
| **4** | `Path.write_text` patches too broad | ✅ Fixed | Narrowed to module-level scope |
| **5** | `test_del_calls_cleanup` wrong contract | ✅ Resolved | Test and `__del__` method both deleted — `CleanupManager` owns the lifecycle now |
| **6** | `KEEP_REPORT_DATA` env var scattered | ⏭️ Deferred to #538 | Pre-existing debt with its own tracking issue |
| **7** | `object.__new__` fragile coupling | ✅ Fixed | Tests rewritten without `__new__` bypass |
| **8** | Missing `_cleanup_subprocess_runner` tests | ✅ Resolved | Method deleted entirely — `CleanupManager` replaces it |
| **9** | Test file placement deepens #541 | ✅ Fixed | Moved `test_orchestrator_config_error.py` to `tests/unit/pyats_core/` |
| **10** | `ArchiveConfigPlugin` credential leak | ⏭️ Tracked as #773 | Low risk (nac-test creds are always env vars), but user-provided testbeds with hardcoded credentials could leak |

The key architectural change: by adopting `CleanupManager` (merged in #727), we eliminated `SubprocessRunner.cleanup()`, `SubprocessRunner.__del__()`, and `PyATSOrchestrator._cleanup_subprocess_runner()` entirely. Config files are registered at creation time and cleaned up automatically — which also resolves the early-return gap from comment #1 by design.

aitestino (approved):

PR #718, opened 2026-03-26
Fix robot command not found when installed via uv tool

danischm:
## Summary
- Use `sys.executable -m robot` instead of bare `robot` for pabot subprocess invocation
- Fixes "robot: command not found" when nac-test is installed in an isolated environment (e.g. `uv tool install`)

## Root cause
Pabot spawns `robot` as a subprocess by name. When nac-test is installed via `uv tool install`, the `robot` binary lives inside the isolated venv and is not on the system PATH.

## Fix
Pass `--command <python> -m robot --end-command` to pabot so it invokes robot through the current Python interpreter, which always resolves to the correct venv.

### DIFF
--- a/nac_test/robot/pabot.py
+++ b/nac_test/robot/pabot.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2025 Daniel Schmidt
 
 import logging
-import sys
+import sysconfig
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


oboehmer (approved):
will push a different fix as discussed.. also raise an issue to properly test this in the pipeline

PR #727, opened 2026-03-27
fix(security): auto-cleanup merged_data_model.yaml, remove CLI option and eliminate redundant re-read (#677 #704)

### LINKED ISSUE #573: Add logger.exception() to main.py exception handler for full stack traces
## Context

In `nac_test/cli/main.py` lines 374-377, the exception handler suppresses the exception chain in production mode (`DEBUG_MODE=False`):

```python
except Exception as e:
    typer.echo(f"Error during execution: {e}")
    if DEBUG_MODE:
        raise typer.Exit(EXIT_ERROR) from e
    raise typer.Exit(EXIT_ERROR) from None  # ← Chain suppressed
```

## Issue

The `from None` suppresses the full exception chain. For chained exceptions (e.g., `ConnectionError -> RuntimeError`), only the top-level message is visible.

## Recommendation

Add `logger.exception()` to preserve full context in logs:

```python
except Exception as e:
    logger.exception("Unexpected error during execution")  # ← Add this
    typer.echo(f"\n❌ Error during execution: {e}", err=True)
    if DEBUG_MODE:
        raise typer.Exit(EXIT_ERROR) from e
    raise typer.Exit(EXIT_ERROR) from None
```

This preserves exception chains in logs while keeping terminal output clean for users.

## Related

PR #553

### LINKED ISSUE #640: refactor: introduce DEFAULT_MERGED_DATA_FILENAME constant as single source of truth
## Problem

The string \`"merged_data_model_test_variables.yaml"\` is used as a hardcoded literal in multiple Python source files:

- \`nac_test/cli/main.py\` — CLI default value
- \`nac_test/data_merger.py\` — \`write_merged_data_model()\` default arg
- \`nac_test/robot/robot_writer.py\` — \`write_merged_data_model()\` default arg
- \`nac_test/pyats_core/execution/device/device_executor.py\` — hardcoded path construction
- \`tests/e2e/test_e2e_scenarios.py\` — output-root whitelist assertion
- \`tests/integration/test_cli_rendering.py\` — expected filename assertion

Renaming the file requires grep-and-replace across all these sites with no compiler safety net.

## Proposed fix

1. Add to \`nac_test/core/constants.py\`:
   \`\`\`python
   DEFAULT_MERGED_DATA_FILENAME: str = "merged_data_model_test_variables.yaml"
   \`\`\`
2. Replace all Python source occurrences of the raw string with the constant (docs/markdown references can stay as-is — they are user-facing prose, not code).

## Files to update (Python only)

- \`nac_test/core/constants.py\` — add constant
- \`nac_test/cli/main.py\`
- \`nac_test/data_merger.py\`
- \`nac_test/robot/robot_writer.py\`
- \`nac_test/pyats_core/execution/device/device_executor.py\`
- \`tests/e2e/test_e2e_scenarios.py\`
- \`tests/integration/test_cli_rendering.py\`

### LINKED ISSUE #677: Remove merged data model artifact after test execution to prevent potential credential exposure

### LINKED ISSUE #704: perf: Eliminate duplicate data merging in RobotWriter
### Problem

Data files are currently merged twice during test execution:

1. **CLI (`main.py`)** — merges data and writes `merged_data_model_test_variables.yaml`
2. **RobotWriter** — re-merges the same files when initialized

### Current Flow

CLI (lines 398-399):
```python
merged_data = DataMerger.merge_data_files(data)
DataMerger.write_merged_data_model(merged_data, output, merged_data_filename)
```

RobotWriter `__init__` (line 70):
```python
self.data = DataMerger.merge_data_files(data_paths)
```

RobotOrchestrator then calls `write_merged_data_model()` which writes the file again.

### Proposed Solution

Refactor RobotWriter to accept the merged data path instead of raw data_paths, and read the file that CLI already created:

```python
# Current
RobotWriter(data_paths: list[Path], ...)

# Proposed  
RobotWriter(merged_data_path: Path, ...)
```

This aligns with how PyATS orchestrator handles merged data.

### Related

Identified during #677 work.

### LINKED ISSUE #717: test: add unit and integration test coverage for DataMerger
## Summary

`DataMerger` has no dedicated test coverage. It is only referenced in existing tests as a mock target to prevent real YAML merging from running. The three static methods — `merge_data_files`, `write_merged_data_model`, and `load_yaml_file` — are untested at both the unit and integration level.

### Unit tests (`tests/unit/test_data_merger.py`)

Cover the wrapper-level behaviour that `DataMerger` owns, mocking `nac_yaml.yaml` to keep tests fast and focused:

- `merge_data_files` returns a `dict` when `yaml.load_yaml_files` returns a valid dict
- `merge_data_files` returns `{}` when `yaml.load_yaml_files` returns `None`
- `merge_data_files` returns `{}` when `yaml.load_yaml_files` returns a non-dict (e.g. a list)
- `write_merged_data_model` calls `yaml.write_yaml_file` with the expected path (`output_directory / filename`)
- `write_merged_data_model` uses the default filename when none is supplied
- `load_yaml_file` wraps the single-path call correctly (passes `[file_path]` to `yaml.load_yaml_files`)
- `load_yaml_file` returns `{}` when `yaml.load_yaml_files` returns `None`

### Integration tests (`tests/integration/test_data_merger_integration.py`)

Exercise real YAML loading with `tmp_path` fixtures — no mocking of `nac_yaml`. The goal is to catch regressions in the full merge pipeline as seen by nac-test, even though recursive directory walking is `nac_yaml`'s responsibility.

Proposed scenarios (YAML fixtures can be placed under `tests/integration/fixtures/data_merger/`):

| Scenario | Description |
|---|---|
| Single flat file | One file loaded, correct dict returned |
| Multiple flat files | Two files merged, keys combined |
| Flat directory | Directory passed; all `.yaml` files inside are picked up |
| Recursive directory | Subdirectory tree; all `.yaml` files found recursively |
| Mixed files and dirs | List contains both explicit file paths and directory paths |
| Empty directory | No YAML files; returns `{}` |
| Empty/null YAML | File exists but is empty or contains only `null`; returns `{}` |
| Write round-trip | `write_merged_data_model` writes a file that `load_yaml_file` reads back correctly |

The recursive-directory scenario is the most important: even though walking is delegated to `nac_yaml`, having an integration test here ensures the contract (pass a `Path` directory → get merged result) is visible in nac-test's own test suite and will catch silent breakage if the dependency changes.

## Files to create

- `tests/integration/test_data_merger_integration.py`
- `tests/unit/test_data_merger.py`
- `tests/integration/fixtures/data_merger/` — YAML fixtures for integration tests

## Out of scope

- Deduplication logic (owned and tested by `nac_yaml`)
- Merge conflict resolution strategies (owned by `nac_yaml`)

oboehmer:
## Description

`merged_data_model_test_variables.yaml` is written to the output directory on every run and contains the full merged data model, which may include credentials and other sensitive values. Prior to this PR the file was never deleted, leaving it on disk after test execution. This PR registers it with `CleanupManager` for automatic deletion on normal exit, SIGTERM, and SIGINT.

Per-device job scripts (`.py`) and testbed files (`.yaml`) created by `DeviceExecutor` with `NamedTemporaryFile(delete=False)` were similarly never deleted, accumulating in the OS temp directory across runs. These are also now registered with `CleanupManager`.

As part of the same cleanup, a performance regression is also fixed: `RobotWriter` was re-reading and re-parsing the merged data YAML from disk after it had just been written — costing ~51 s and ~400 MB peak RAM on large data models. The already-loaded dict is now passed through the call chain (`CombinedOrchestrator` → `RobotOrchestrator` → `RobotWriter`) directly, eliminating the redundant file read entirely.

Additional simplifications made possible by these changes:

- **`RobotWriter._convert_data_model_for_templates()` deleted** — this method performed a JSON round-trip (`json.dumps` → `json.loads`) to convert `CommentedMap` (ruamel.yaml's dict subclass) into plain Python dicts before passing data to Jinja2. This was originally necessary to avoid duplicate-key issues with older ruamel.yaml behaviour. It is now safe to remove because **nac-yaml 2.0 guarantees plain `dict` / `list` output** — `CommentedMap` and `CommentedSeq` are fully unwrapped before being returned to callers. Jinja2 receives ordinary Python types directly, so no conversion layer is needed. Removing the JSON round-trip is also the primary source of the ~51 s / ~400 MB saving on large data models.
- **`DataMerger.load_yaml_file()` removed** — no remaining callers.
- **`DataMerger.write_merged_data_model()` `filename` parameter removed** — the output filename is now always `MERGED_DATA_FILENAME` (renamed from `DEFAULT_MERGED_DATA_FILENAME`; the `DEFAULT_` prefix implied variability that no longer exists).
- **`DataMerger.merged_data_path()`** added as single query point for the file location — callers no longer reconstruct `output_dir / MERGED_DATA_FILENAME` independently.
- **`write_merged_data_model()` returns the `Path` it wrote** and now owns the `chmod 0o600` call — "write this sensitive file securely" is a single responsibility.
- **`DeviceExecutor`** receives `merged_data_path` as a constructor argument from the orchestrator (which already holds it) rather than re-deriving it from `base_output_dir`, removing the `DataMerger` import from `device_executor.py` entirely.
- **`CleanupManager.register()`** gains a `keep_if_debug` flag. Files registered with `keep_if_debug=True` are retained when `NAC_TEST_DEBUG` is set, letting developers inspect intermediate artifacts (merged data YAML, job scripts, testbed YAMLs) without any extra flags or code changes.
- **`logger.exception()` added** to the top-level exception handler in `main()` — the full stack trace (including chained causes) is now captured in the log file while terminal output remains a clean single line for end users ( unrelated to this PR, but I touched this area so I fixed #573 while at it)
- **`MERGED_DATA_FILENAME` constant** introduced as the single source of truth for the merged data filename, replacing hardcoded string literals across all call sites.
- **`DataMerger` unit and integration tests** added (`tests/unit/test_data_merger.py`).

## Closes

- Fixes #573
- Fixes #640
- Fixes #677
- Fixes #704
- Fixes #717

## Related Issue(s)

- N/A

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Breaking Changes

The `--merged-data-filename` CLI option has been removed. The output filename for the merged data model is now always `merged_data_model_test_variables.yaml` and cannot be customised. Any scripts or CI pipelines referencing `--merged-data-filename` must remove that argument.

## Test Framework Affected

- [ ] PyATS
- [x] Robot Framework
- [ ] Both

- [x] All architectures

## Platform Tested

- [x] macOS (version tested: macOS 15 / Python 3.12)

## Key Changes

- `CleanupManager`: new utility that deletes registered files on exit, SIGTERM, and SIGINT; `register()` gains `keep_if_debug` flag
- `main.py`: registers `merged_data_model_test_variables.yaml` with `CleanupManager` immediately after writing; `--merged-data-filename` CLI option removed; `exit_with_cleanup()` closure removed;
- `DeviceExecutor`: registers per-device job scripts and testbed YAMLs with `CleanupManager`; receives `merged_data_path` injected by orchestrator rather than re-deriving it
- `RobotWriter`: constructor now accepts `merged_data: dict` instead of `merged_data_path: Path`; no file I/O or type conversion on init
- `RobotOrchestrator` / `CombinedOrchestrator`: accept and forward `merged_data` dict
- `DataMerger`: removes `filename` parameter from `write_merged_data_model()`, removes `load_yaml_file()` entirely, adds `merged_data_path()` helper, returns `Path` from `write_merged_data_model()`, owns `chmod 0o600`
- `_convert_data_model_for_templates()` deleted — safe due to nac-yaml 2.0 plain-dict output guarantee
- `MERGED_DATA_FILE_MODE = 0o600` constant added; `DEFAULT_MERGED_DATA_FILENAME` renamed to `MERGED_DATA_FILENAME`

## Testing Done

- [x] Integration tests added/updated

### Test Commands Used

```bash
uv run pytest -n auto --dist loadscope tests/
uv run pre-commit run -a
```

## Additional Notes

PyATS subprocesses still read the merged data file from disk during their execution window — the file is only deleted after all test execution completes, so this is safe.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aitestino (changes requested):
Hey @oboehmer, thank you for the PR -- this is a really solid piece of work that addresses a genuine security gap (credentials lingering on disk in `merged_data_model_test_variables.yaml`) while simultaneously delivering a meaningful performance win (~51s / ~400MB on large data models). The architectural choices here are all correct: threading the `merged_data` dict through the call chain instead of re-reading from disk is textbook dependency injection, the `CleanupManager` singleton is the right pattern for process-wide exit cleanup, and centralizing the filename into `MERGED_DATA_FILENAME` with `DataMerger.merged_data_path()` as the single source of truth is a clean DRY improvement. The removal of `_convert_data_model_for_templates()` leveraging nac-yaml 2.0's plain-dict guarantee is a nice simplification, and the test coverage across unit, integration (real subprocess signal delivery), and E2E is thorough.

I did a couple of thorough reviews on this, and I think we need a few changes before merge. The core design is sound -- most of these are correctness and polish items.

1. **Signal handler deadlock on `threading.Lock`** -- `_signal_handler()` calls `run_cleanup()`, which acquires `self._lock`. If a SIGTERM or SIGINT is delivered while the main thread is already inside `register()` or `unregister()` (which also hold `self._lock`), this will deadlock -- Python's `threading.Lock` is not reentrant, and signal handlers run in the main thread between bytecodes. The fix is straightforward: replace `threading.Lock()` with `threading.RLock()` in `__init__`. One-line change. This is the only item I'd consider blocking.

2. **`keep_if_debug=True` on the merged data file** -- The merged data file can contain credentials (the e2e fixture shows `test_credential: !env TEST_CREDENTIAL_SENTINEL`). Registering it with `keep_if_debug=True` means it persists when `NAC_TEST_DEBUG=true` -- which CI/CD pipelines might set inadvertently, leaving credentials in build artifacts. From a best practices standpoint, I think ideally we want to register the merged data file *without* `keep_if_debug=True` since it's the one file that's explicitly sensitive. Debug users can still inspect it during execution; it only needs to survive until cleanup runs. What's your thinking here?

3. **Duplicated `merged_data` defaulting pattern** -- `merged_data if merged_data is not None else {}` appears identically in both `CombinedOrchestrator.__init__()` and `RobotOrchestrator.__init__()`. A small helper would close this:
   ```python
   def coerce_merged_data(merged_data: dict[str, Any] | None) -> dict[str, Any]:
       return merged_data if merged_data is not None else {}
   ```

4. **Post-cleanup `register()` calls are silently lost** -- Once `run_cleanup()` sets `_cleanup_done = True`, any file registered afterward will never be deleted (subsequent `run_cleanup()` calls are no-ops). The thread safety test actually demonstrates this race and works around it by manually resetting `_cleanup_done = False`. If this is an accepted product behavior (files registered after cleanup starts are permanently lost), a brief note in the `register()` docstring would make the contract explicit. If it's unintended, the fix would be to have late registrations trigger immediate deletion.

5. **`test_run_tests_full_execution` doesn't use the new `_setup_run_tests_mocks` helper** -- The PR introduces `_setup_run_tests_mocks` specifically to eliminate repeated mock wiring, and two sibling tests use it. `test_run_tests_full_execution` still inlines the same setup manually. Having it delegate to the helper would be consistent.

6. **Missing inline comment for `_convert_data_model_for_templates()` removal** -- The JSON round-trip removal relies on nac-yaml 2.0 guaranteeing plain `dict`/`list` output. A one-line comment in `RobotWriter.__init__` noting this dependency (e.g., `# nac-yaml 2.0+ guarantees plain dict/list -- no OrderedDict conversion needed`) would save future developers from wondering why the conversion was removed.

7. **PR description mentions fixture move that didn't happen** -- "Data merge fixtures moved from `tests/integration/fixtures/` to `tests/unit/fixtures/`" -- this move isn't in the diff. Minor documentation accuracy fix.

I'd consider item 1 as the one thing to address before merge -- it's a one-liner. Item 2 is worth a conversation since it's a security posture decision. Items 3-7 are lower priority and could be follow-ups if you'd rather not hold this up further.

What do you think?

_P.S. -- This comment was drafted using voice-to-text via Claude Code. If the tone comes across as overly direct or terse, please know that's just how it tends to phrase things. No offense or criticism is intended -- this is purely an objective technical review of the PR. Thanks for understanding! 🙂_

oboehmer:
Thanks for the thorough review, @aitestino — all good catches. Here's what I've done:

**1. Signal handler deadlock (RLock)** — Agreed, fixed in 6501961. `threading.Lock()` → `threading.RLock()`.

**2. `keep_if_debug=True` on merged data** — Good call on the security angle. As I felt some pushback on the removal in a prior discussion with you, I explicitly added this to address your concerns — but you're right, the merged data file is the one file that's explicitly sensitive. Removed in 8e103d4 — merged data is now always cleaned up regardless of `NAC_TEST_DEBUG`.

**3. Duplicated `merged_data` defaulting** — I'll push back gently on this one. These are two identical one-liners (`merged_data if merged_data is not None else {}`) in two separate classes with no shared base. Extracting a helper for a single `or`-style expression adds indirection without meaningful DRY benefit — the "duplication" is simpler than the abstraction. If a third caller appears or the logic grows, happy to revisit.

**4. Post-cleanup `register()` calls silently lost** — This was indeed a bug, not intended behavior. The root cause was `_cleanup_done` — once set, it made `run_cleanup()` a permanent no-op, so files registered between signal-handler cleanup and atexit cleanup were never deleted. The fix (fc73d12) removes `_cleanup_done` entirely: `_files.clear()` at the end of `run_cleanup()` already provides natural idempotency (iterating an empty dict is a no-op), and new files registered after a cleanup are properly picked up by the next invocation. Added a dedicated test (`test_files_registered_between_cleanups_are_deleted`) that validates the full register → cleanup → register → cleanup cycle.

**5. `test_run_tests_full_execution` not using helper** — Agreed, refactored in 2d73e98.

**6. Missing nac-yaml 2.0 comment** — Added in 137c7e6.

**7. PR description fixture move** — Removed the incorrect line.

oboehmer:
Merging this now as all comments have been addressed and you're on holiday. happy to address changes in follow-up PRs.

PR #467, opened 2026-01-28
Fix ConnectionBroker Event Loop Issues and Connection Pooling

oboehmer:
## Summary

This PR resolves critical regressions preventing the ConnectionBroker from functioning properly:
1. Event loop management bug causing RuntimeError crashes
2. Connection pooling being bypassed, resulting in duplicate connections
3. Connections being disconnected after each test, defeating pooling benefits

After iterative investigation, we discovered the broker had three compounding issues that made it completely non-functional.

### 1. Broker Bypassed (Issue #464 - Connection Pooling Not Working) - REGRESSION

Problem: Broker daemon was running but never used - tests created direct connections instead.

Regression introduced: When testbed parameter support was added to PyATS job generation in a later commit, the testbed became available in test context, which changed the connection priority logic.

The connection logic checked for testbed device BEFORE checking for broker availability. When testbed was present (needed for Genie parsers), the testbed path was taken, bypassing the broker entirely.

Fix: Check NAC_TEST_BROKER_SOCKET environment variable FIRST to prioritize broker when active, regardless of testbed availability.

### 2. Event Loop Error (Issue #465 - Tests Failing) - REGRESSION

Problem: Tests crashed with RuntimeError: Task got Future attached to a different loop

Regression introduced: I missed one instance where get_or_create_event_loop() should have been used when refactoring event loop management. noticed this once the connection broker started working with fix above

The code attempted to call run_until_complete() on an already-running event loop, which is invalid.

Fix: Use get_or_create_event_loop() utility consistently, as done elsewhere in the codebase.

### 3. Connections Disconnected After Each Test (Discovered During Testing)

Problem: Even with broker active, connections were being explicitly disconnected in test cleanup, clearing the connection pool and command cache between tests.

Root cause: Original broker implementation did not account for test cleanup hooks that disconnect all connections. This was always present but only became apparent when testing the fixed broker.

Fix: Skip disconnect() calls for BrokerCommandExecutor connections - the broker manages lifecycle.

### ConnectionBroker Statistics Tracking
- Added: Minimal statistics collection (connection/command cache hits/misses)
- Why: Enable validation that broker is actually working as intended
- How: Log structured statistics during broker shutdown for test parsing
- Result: 75% connection reuse, 67% command cache hit rate in tests

### SSHTestBase Connection Management
- Changed: Connection setup now prioritizes broker over testbed
- Changed: Cleanup skips disconnect for broker-managed connections
- Changed: Added type annotations for mypy compatibility
[...]

PR #732, opened 2026-03-30
fix(broker): clean up error output, add failure-mode tests, harden socket validation (#539, #486, #485)

### LINKED ISSUE #485: [ConnectionBroker] Add socket validation and graceful fallback
## Summary

The ConnectionBroker currently assumes the broker is running if `NAC_TEST_BROKER_SOCKET` environment variable is set. There's no validation that the socket actually exists or that the broker is listening.

## Current Behavior

```python
broker_active = "NAC_TEST_BROKER_SOCKET" in os.environ
if broker_active:
    # Assumes broker is running, attempts connection
```

### 1. Socket Validation
Before attempting to use the broker, validate:
- Socket file exists at the specified path
- Broker is actually listening (optional: connection health check)

```python
broker_socket = os.environ.get("NAC_TEST_BROKER_SOCKET")
if broker_socket:
    socket_path = Path(broker_socket)
    if not socket_path.exists():
        logger.warning(f"Broker socket not found: {socket_path}, falling back to direct connection")
        broker_active = False
    else:
        broker_active = True
```

### 2. Graceful Fallback
If broker connection fails, fall back to direct testbed connection instead of crashing:
- Catch broker connection errors
- Log warning about fallback
- Continue with direct connection path

### 3. Resource Leak Mitigation
Currently, if the broker crashes mid-test, connections managed by it won't be cleaned up (tests skip `disconnect()` for broker connections).

Consider:
- Add fallback cleanup logic if broker is unreachable during test cleanup
- Document the resource leak risk for production deployments

## Related

- PR #467 introduced the broker prioritization logic
- This is hardening work to make the broker more robust

### LINKED ISSUE #486: [ConnectionBroker] Add failure mode test coverage
## Summary

The ConnectionBroker integration tests validate the happy path (connection pooling works, command caching works), but don't cover failure scenarios that could occur in production.

## Current Test Coverage

✅ Connection pooling works when broker is running
✅ Command caching returns cached results  
✅ Statistics are logged correctly
✅ Detection of broker bypass scenarios

### 1. Broker Unavailable Scenarios
- `NAC_TEST_BROKER_SOCKET` set but broker not running
- Socket file exists but broker process crashed
- Socket path is invalid/garbage

### 2. Broker Crash Mid-Test
- Broker dies during test execution
- Socket deleted while tests running
- Connection to broker times out

### 3. Connection Failures
- Network device disconnect fails
- Broker connection health check fails
- Command execution timeout

### 4. Concurrent Access
- Multiple test processes accessing broker simultaneously
- Race conditions in connection pool management

### 5. Resource Cleanup
- Validate cleanup when broker crashes
- Memory leak testing with large test suites (1000+ tests)

## Proposed Test Strategy

1. **Unit tests** for validation logic (socket exists, broker reachable)
2. **Integration tests** with mock broker that can simulate failures
3. **Chaos testing** - kill broker process during test run, verify graceful handling

## Related

- PR #467 added broker integration tests for happy path
- Issue #485 tracks socket validation and graceful fallback implementation

### LINKED ISSUE #539: fix(broker): improve error reporting when device connection fails via ConnectionBroker
## Context

After PR #467 landed (which fixed the broker being bypassed — great catch on the priority ordering), I noticed a change in how connection failures are surfaced to the user during D2D test execution.

When a device is unreachable (e.g., Unicon times out trying to reach `iosxe-r1`), the output now includes the full Unicon state machine exception chain — including every regex pattern Unicon uses for prompt detection — followed by a generic "Unknown broker error" message. The result is a wall of text that makes it difficult to identify the actual failure reason.

This wasn't visible before #467 because the broker path was being bypassed (testbed was checked first), so the `_create_connection` error handler in `connection_broker.py` never fired during connection failures.

## Observed Behavior

When a device connection fails through the broker, the user sees:

1. **Full Unicon traceback with all prompt regex patterns** (~100+ lines) — caused by `exc_info=True` at `connection_broker.py:_create_connection`
2. **"Broker error: Unknown broker error"** — because the `"connect"` command response returns `{"status": "error", "result": false}` without an `"error"` key, so `broker_client.py` falls back to the default message
3. **The same failure logged 4 times** at different levels of the broker → client → test call chain

## Expected Behavior

A concise, actionable error message like:

```
ERROR - ConnectionBroker: failed to connect to iosxe-r1: TimeoutError (60s)
FAILED operational.verify_bgp_peers - Device connection failed: iosxe-r1
```

## Proposed Changes

All three are small, scoped fixes in `connection_broker.py` and `broker_client.py`:

### 1. Replace `exc_info=True` with concise logging

In `_create_connection` (~line 340):

```python
# Before
logger.error(f"Failed to connect to {hostname}: {e}", exc_info=True)

# After
logger.error(f"Failed to connect to {hostname}: {e}")
```

The full Unicon traceback is already written to the per-device Unicon log file (`{hostname}-cli-{timestamp}.log`), so it's available for debugging without flooding stdout.

### 2. Include the actual error in the "connect" command response

In `_process_request`, the `"connect"` handler (~line 201):

```python
# Before
success = await self._ensure_connection(hostname)
return {"status": "success" if success else "error", "result": success}

# After — propagate the actual error message
try:
    success = await self._ensure_connection(hostname)
    return {"status": "success" if success else "error", "result": success}
except Exception as e:
    return {"status": "error", "error": str(e)}
```

*Note: this may require `_ensure_connection` to raise instead of returning `False`, or `_get_connection`/`_create_connection` to propagate the exception — the exact approach is flexible.*

### 3. Reduce cascading error logs

The broker should log once at the source (`_create_connection`). The intermediate layers (`_send_request`, `ensure_connection`) should log at `debug` level or propagate silently, so the test layer can emit the single user-facing error.

## References

- Introduced by: #467 (`6b344454`)
- Specifically: the priority reorder in `ssh_base_test.py:_async_setup` that activated the broker path
- The `exc_info=True` was present since the original broker implementation (Sept 2025, `0d5cd6ad`) but never triggered until #467

## Terminal Output Showcasing This

```bash
nac-sdwan-terraform on  review-branch [$?] via 🐍 v3.12.9 (.venv) via 💎 v3.0.0 via 💠 default 
❯ nac-test -d ./data -t ./tests/ -o ./nac-test-results --pyats


[18:44:23] 📄 Merging data model files...
[18:44:23] ✅ Data model merging completed (0.3s)


⚠️  WARNING: --pyats flag is for development use only. Production runs should use combined execution.
🧪 Running PyATS tests only (development mode)...
Discovered 3 PyATS test files
Running with 1 parallel workers
Found 1 API test(s) - using standard execution

WARNING - Skipping device C8K-40C0CCFD-9EA8-2B2E-E73B-32C5924EC79B: management_ip_variable 'vpn10_Lo10_if_ipv4_address' not found in device_variables.
WARNING - Skipping device C8K-E94D7B88-4B9E-3323-C6C3-F29079FAAC3B: management_ip_variable 'vpn10_Lo10_if_ipv4_address' not found in device_variables.
WARNING - Skipping device C8K-B5EE117E-1210-C396-EC41-D3DA64D6A51A: management_ip_variable 'vpn10_Lo10_if_ipv4_address' not found in device_variables.
WARNING - Skipping device C8K-83E4B5EE-52AC-075F-B5E4-0050110680F3: management_ip_variable 'vpn10_Lo10_if_ipv4_address' not found in device_variables.
WARNING - Skipping device C8K-D4D8222C-26DA-83D1-EFA8-EAA6C881AFBF: management_ip_variable 'vpn10_Lo10_if_ipv4_address' not found in device_variables.
WARNING - Skipping device C8K-25679DF5-DE95-81BC-5D4A-56907294945B: management_ip_variable 'vpn10_Lo10_if_ipv4_address' not found in device_variables.

Found 2 D2D test(s) - using device-centric execution

Executing PyATS with command: /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/.venv/bin/pyats run job /var/folders/75/qp72b5rn3bj856jfnql062hr0000gn/T/tmpf765mcpv_api_job.py --configuration /var/folders/75/qp72b5rn3bj856jfnql062hr0000gn/T/tmpezscd6sr_plugin_config.yaml --archive-dir /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results --archive-name nac_test_job_20260216_184423_800.zip --no-archive-subdir --no-mail --no-xml-report

Executing PyATS with command: /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/.venv/bin/pyats run job /var/folders/75/qp72b5rn3bj856jfnql062hr0000gn/T/tmpxvlmnc0f.py --testbed-file /var/folders/75/qp72b5rn3bj856jfnql062hr0000gn/T/tmpr9_g5mi6.yaml --configuration /var/folders/75/qp72b5rn3bj856jfnql062hr0000gn/T/tmp5eq7mhzs_plugin_config.yaml --archive-dir /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results --archive-name pyats_archive_device_iosxe-r1 --no-archive-subdir --no-mail --no-xml-report --quiet
2026-02-16 18:44:34.623 [PID:77151] [76945] [ID:1] EXECUTING nrfu.verify_sdwanmanager_device_status
2026-02-16 18:44:34.715 [PID:77154] [76947] [ID:2] EXECUTING operational.verify_bgp_peers
2026-02-16 18:44:36.228 [PID:77151] [76945] [ID:1] PASSED nrfu.verify_sdwanmanager_device_status in 1.6 seconds
------------------------------------------------------------------------------
Verify All Managed Devices Are Normal                                  | PASSED |
------------------------------------------------------------------------------
ERROR - Failed to connect to iosxe-r1: failed to connect to iosxe-r1
Failed while bringing device to "any" state
Traceback (most recent call last):
  File "src/unicon/statemachine/statemachine.py", line 738, in unicon.statemachine.statemachine.StateMachine.go_to
  File "src/unicon/statemachine/statetransition.py", line 480, in unicon.statemachine.statetransition.AnyStateTransition.do_transitions
  File "src/unicon/eal/dialogs.py", line 479, in unicon.eal.dialogs.Dialog.process
  File "src/unicon/eal/dialog_processor.py", line 358, in unicon.eal.dialog_processor.SimpleDialogProcessor.process
  File "src/unicon/eal/dialog_processor.py", line 304, in unicon.eal.dialog_processor.SimpleDialogProcessor.timeout_handler
unicon.core.errors.TimeoutError: timeout occurred:
              timeout value: 60
              last_command: '\r'
             pattern: ['^.*--\\s?[Mm]ore\\s?--.*$', '^.*?% (Bad passwords|Bad secrets|Access denied|Authentication failed|Login invalid).*?\\n', '^.*Login incorrect', '^.*% (No password set|Error in authentication.).*', '^.*([Uu]sername|(?<!Last )[Ll]ogin): ?$', '^.*User Access Verification', '^(Enter new password|Confirm password):\\s*$', '^.*[Pp]assword( for )?(\\S+)?: ?$', '^(.*)Kerberos:\\s*No default realm defined for Kerberos!\\s*$', 'Password OK\\s*$', '^.*Enter passphrase for key .*?:\\s*?', '^.*?(Enter|Confirm) enable secret( \\[<Use current secret>\\])?:\\s*$', '^.*(initial|basic) configuration dialog\\s?\\??\\s*((\\[yes.*\\])|(\\(yes/no\\))|(\\(yes/no\\)\\s*(\\[y\\]))):', 'Would you like to enter basic management setup\\? \\[yes/no\\]:\\s*$', '^.*?Enter your selection( \\[2])?:\\s*$', '^.*?Enter your encryption selection( \\[2])?:\\s*$', '^\\+\\>\\s?$', 'Escape character is .*\\n', '^.*RETURN to get started', 'Are you sure you want to continue connecting \\(yes/no(/\\[fingerprint\\])?\\)', '^.*(Connection refused|Connection reset by peer|Broken pipe|Network is down|closed by remote host|Connection closed by foreign host)', 'Received disconnect from .*:', 'Hit Enter to proceed:', '^(.*?)Press Ctrl\\+x to Exit the session', '^(.*?)Connected.', '^.*?(%\\w+(-\\S+)?-\\d+-\\w+|yang-infra:|PKI_SSL_IPC:|Guestshell destroyed successfully|%Error opening tftp:\\/\\/255\\.255\\.255\\.255|Autoinstall trying|audit: kauditd hold queue overflow|SECURITY WARNING|%RSA key|INSECURE DYNAMIC WARNING|(LC|RP)/\\d+/\\d+/CPU\\d+:\\w+\\s+\\d+\\s+\\d{2}:\\d{2}:\\d{2}|\\[OK\\]).*\\s*$', '^.*?Press any key to continue\\..*?$', '^.*[Pp]ermission denied', '^.*?enable[\\r\\n]*.*?[Pp]assword( for )?(\\S+)?: ?$', '^.*?(%\\w+(-\\S+)?-\\d+-\\w+|yang-infra:|PKI_SSL_IPC:|Guestshell destroyed successfully|%Error opening tftp:\\/\\/255\\.255\\.255\\.255|Autoinstall trying|audit: kauditd hold queue overflow|SECURITY WARNING|%RSA key|INSECURE DYNAMIC WARNING|(LC|RP)/\\d+/\\d+/CPU\\d+:\\w+\\s+\\d+\\s+\\d{2}:\\d{2}:\\d{2}|\\[OK\\]).*\\s*$', 'Uncommitted changes found, commit them\\? \\[yes\\/no\\/CANCEL\\]\\s*$', '^.*System config(uration)? has been modified\\.\\s*Save\\?.*$', '^.*Are you sure you want to reset the system\\s*\\(y\\/n\\)\\?', '^.*Reload node\\s*\\?\\s*\\[no,yes\\]\\s?$', '^.*Proceed( with( quick)? reload)?\\?\\s*\\[confirm\\]', '^.*Do you wish to proceed with reload anyway\\s*\\[confirm\\]\\s*', '^.*User Access Verification', '^(.*)Uncommitted changes found.*?', '^.*(initial|basic) configuration dialog\\s?\\??\\s*((\\[yes.*\\])|(\\(yes/no\\))|(\\(yes/no\\)\\s*(\\[y\\]))):', '^(.*)Would you like to terminate autoinstall\\?\\s*\\[yes\\]:\\s*$', '^.*Do you want to reload the internal AP\\s?\\?\\s((\\[yes/no\\]\\??)|(\\[y/n\\]\\??)):?\\s?$', '^.*Do you want to save the configuration of the AP\\s?\\?\\s*((\\[yes/no\\]\\??)|(\\[y/n\\]\\??)):?\\s?$', '^(.*)This command will reboot the system.\\s*\\(y\\/n\\)\\?\\s*\\[n\\]\\s?$', '^.*Do you want to enforce secure password standard(\\?)?\\s*\\(yes\\/no\\)(\\s*\\[[yn]\\])?\\:\\s*', '^.*(Enter|Confirm) the password for .*admin', 'Abort( Power On)? Auto Provisioning .*:', '^.*([Uu]sername|(?<!Last )[Ll]ogin): ?$', '^.*[Pp]assword( for )?(\\S+)?: ?$', '^.*Press RETURN to get started.*', 'Press RETURN to get started.*', '^(.*?)Connection.*? closed|disconnect: Broken pipe', '^(.*)Please reset', '^.*(Enter|Confirm) the password for .*admin', '^.*(initial|basic) configuration dialog.*\\s?', '^(.*)Would you like to terminate autoinstall\\? ?\\[yes\\]: $', '__timeout__', '.*The highlighted entry will be (booted|executed) automatically in .*?(\\x1b\\S+)?\\s+', '^.*Are you sure you want to continue\\? \\[y/n\\]\\s?.*$', '^(.*?)\\[((?P<hostname00>([^# \\t\\n\\r\\f\\v\\(\\)]+))|[Ss]witch|[Rr]outer|eWLC).*?\\]\\$\\s?$', '^(.*?)(rommon[\\s\\d]*>|switch:|grub>)\\s*(\\x1b\\S+)?$', '^(.*)\\(config-pki-hexmode\\)#\\s?$', '^(.*?)(\\(unlicensed\\))?(wlc|WLC|Router|RouterRP|Switch|ios|switch|(?P<hostname00>([^# \\t\\n\\r\\f\\v\\(\\)]+)))([0-9])?(\\(standby\\))?(-stby)?(-standby)?(\\(boot\\))?\\(maint-mode\\)#[\\s\\x07]*$', '^(.*?)(\\{\\.\\.\\}|then.else.fi)\\s*>\\s*$', '^(.*?)\\(rules.*?\\)#[\\s\\x07]*$', '^(.*?)\\(syntax.*?\\)#[\\s\\x07]*$', '^(.*?)\\(acm.*?\\)#[\\s\\x07]*$', '^(.*?)\\(tcl.*?\\)#[\\s\\x07]*$', '^(.*)\\[\\S+@guestshell\\s+.*\\][#\\$]\\s?$', '^(.*)\\((?!.*pki-hexmode).*(con|cfg|ipsec-profile|ca-trustpoint|ca-certificate-map|cs-server|ca-profile|gkm-local-server|cloud|host-list|config-gkm-group|gkm-sa-ipsec|gdoi-coop-ks-config|wsma|enforce-rule|DDNS|ca-trustpool|cert-trustpool)\\S*\\)#\\s?$', '^(.*?)(\\(unlicensed\\))?(wlc|WLC|Router|RouterRP|Switch|ios|switch|(?P<hostname00>([^# \\t\\n\\r\\f\\v\\(\\)]+)))([0-9])?(\\(recovery-mode\\))?(\\(rp-rec-mode\\))?(\\(standby\\))?(-stby)?(-standby)?(\\(boot\\))?#[\\s\\x07]*$', '^(?!.*?grub)(.*?)(\\(unlicensed\\))?(wlc|WLC|Router|RouterRP|Switch|ios|switch|(?P<hostname00>([^# \\t\\n\\r\\f\\v\\(\\)]+)))([0-9])?(\\(recovery-mode\\))?(\\(rp-rec-mode\\))?(\\(standby\\))?(-stby)?(-standby)?(\\(boot\\))?(?<! -)>\\s?$', '^.*?([S|s]tandby console disabled|This \\(D\\)RP Node is not ready or active for login \\/configuration.*)', '^.*--\\s?[Mm]ore\\s?--.*$']
             buffer:'\r\n'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "src/unicon/bases/connection.py", line 839, in unicon.bases.connection.Connection.connect
  File "src/unicon/bases/routers/connection_provider.py", line 300, in unicon.bases.routers.connection_provider.BaseSingleRpConnectionProvider.connect
  File "src/unicon/bases/routers/connection_provider.py", line 331, in unicon.bases.routers.connection_provider.BaseSingleRpConnectionProvider.establish_connection
  File "src/unicon/statemachine/statemachine.py", line 742, in unicon.statemachine.statemachine.StateMachine.go_to
unicon.core.errors.StateMachineError: Failed while bringing device to "any" state

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/atestini/Desktop/Automation/testing-for-nac/nac-test/nac_test/pyats_core/broker/connection_broker.py", line 331, in _create_connection
    await loop.run_in_executor(
  File "/Users/atestini/.pyenv/versions/3.12.9/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/atestini/Desktop/Automation/testing-for-nac/nac-test/nac_test/pyats_core/broker/connection_broker.py", line 333, in <lambda>
    lambda: device.connect(log_stdout=False, logfile=str(logfile_path)),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/pyats/connections/manager.py", line 511, in pyats.connections.manager.ConnectionManager.connect
  File "src/unicon/bases/connection.py", line 850, in unicon.bases.connection.Connection.connect
unicon.core.errors.ConnectionError: failed to connect to iosxe-r1
Failed while bringing device to "any" state
2026-02-16T18:46:00: %NAC_TEST-ERROR: Error communicating with broker: Broker error: Unknown broker error
2026-02-16T18:46:00: %NAC_TEST-ERROR: Failed to ensure connection to iosxe-r1: Broker error: Unknown broker error
2026-02-16T18:46:00: %SCRIPT-ERROR: Failed to connect to device iosxe-r1: Failed to connect to device: iosxe-r1
2026-02-16 18:46:00.693 [PID:77154] [76947] [ID:2] FAILED operational.verify_bgp_peers in 86.0 seconds
------------------------------------------------------------------------------
Verify BGP Peers Status for All SD-WAN Edges (iosxe-r1)                | FAILED |
------------------------------------------------------------------------------
2026-02-16 18:46:00.725 [PID:78350] [76947] [ID:3] EXECUTING operational.verify_ospf_neighbor_state
ERROR - Failed to connect to iosxe-r1: failed to connect to iosxe-r1
Failed while bringing device to "any" state
Traceback (most recent call last):
  File "src/unicon/statemachine/statemachine.py", line 738, in unicon.statemachine.statemachine.StateMachine.go_to
  File "src/unicon/statemachine/statetransition.py", line 480, in unicon.statemachine.statetransition.AnyStateTransition.do_transitions
  File "src/unicon/eal/dialogs.py", line 479, in unicon.eal.dialogs.Dialog.process
  File "src/unicon/eal/dialog_processor.py", line 358, in unicon.eal.dialog_processor.SimpleDialogProcessor.process
  File "src/unicon/eal/dialog_processor.py", line 304, in unicon.eal.dialog_processor.SimpleDialogProcessor.timeout_handler
unicon.core.errors.TimeoutError: timeout occurred:
              timeout value: 60
              last_command: '\r'
             pattern: ['^.*--\\s?[Mm]ore\\s?--.*$', '^.*?% (Bad passwords|Bad secrets|Access denied|Authentication failed|Login invalid).*?\\n', '^.*Login incorrect', '^.*% (No password set|Error in authentication.).*', '^.*([Uu]sername|(?<!Last )[Ll]ogin): ?$', '^.*User Access Verification', '^(Enter new password|Confirm password):\\s*$', '^.*[Pp]assword( for )?(\\S+)?: ?$', '^(.*)Kerberos:\\s*No default realm defined for Kerberos!\\s*$', 'Password OK\\s*$', '^.*Enter passphrase for key .*?:\\s*?', '^.*?(Enter|Confirm) enable secret( \\[<Use current secret>\\])?:\\s*$', '^.*(initial|basic) configuration dialog\\s?\\??\\s*((\\[yes.*\\])|(\\(yes/no\\))|(\\(yes/no\\)\\s*(\\[y\\]))):', 'Would you like to enter basic management setup\\? \\[yes/no\\]:\\s*$', '^.*?Enter your selection( \\[2])?:\\s*$', '^.*?Enter your encryption selection( \\[2])?:\\s*$', '^\\+\\>\\s?$', 'Escape character is .*\\n', '^.*RETURN to get started', 'Are you sure you want to continue connecting \\(yes/no(/\\[fingerprint\\])?\\)', '^.*(Connection refused|Connection reset by peer|Broken pipe|Network is down|closed by remote host|Connection closed by foreign host)', 'Received disconnect from .*:', 'Hit Enter to proceed:', '^(.*?)Press Ctrl\\+x to Exit the session', '^(.*?)Connected.', '^.*?(%\\w+(-\\S+)?-\\d+-\\w+|yang-infra:|PKI_SSL_IPC:|Guestshell destroyed successfully|%Error opening tftp:\\/\\/255\\.255\\.255\\.255|Autoinstall trying|audit: kauditd hold queue overflow|SECURITY WARNING|%RSA key|INSECURE DYNAMIC WARNING|(LC|RP)/\\d+/\\d+/CPU\\d+:\\w+\\s+\\d+\\s+\\d{2}:\\d{2}:\\d{2}|\\[OK\\]).*\\s*$', '^.*?Press any key to continue\\..*?$', '^.*[Pp]ermission denied', '^.*?enable[\\r\\n]*.*?[Pp]assword( for )?(\\S+)?: ?$', '^.*?(%\\w+(-\\S+)?-\\d+-\\w+|yang-infra:|PKI_SSL_IPC:|Guestshell destroyed successfully|%Error opening tftp:\\/\\/255\\.255\\.255\\.255|Autoinstall trying|audit: kauditd hold queue overflow|SECURITY WARNING|%RSA key|INSECURE DYNAMIC WARNING|(LC|RP)/\\d+/\\d+/CPU\\d+:\\w+\\s+\\d+\\s+\\d{2}:\\d{2}:\\d{2}|\\[OK\\]).*\\s*$', 'Uncommitted changes found, commit them\\? \\[yes\\/no\\/CANCEL\\]\\s*$', '^.*System config(uration)? has been modified\\.\\s*Save\\?.*$', '^.*Are you sure you want to reset the system\\s*\\(y\\/n\\)\\?', '^.*Reload node\\s*\\?\\s*\\[no,yes\\]\\s?$', '^.*Proceed( with( quick)? reload)?\\?\\s*\\[confirm\\]', '^.*Do you wish to proceed with reload anyway\\s*\\[confirm\\]\\s*', '^.*User Access Verification', '^(.*)Uncommitted changes found.*?', '^.*(initial|basic) configuration dialog\\s?\\??\\s*((\\[yes.*\\])|(\\(yes/no\\))|(\\(yes/no\\)\\s*(\\[y\\]))):', '^(.*)Would you like to terminate autoinstall\\?\\s*\\[yes\\]:\\s*$', '^.*Do you want to reload the internal AP\\s?\\?\\s((\\[yes/no\\]\\??)|(\\[y/n\\]\\??)):?\\s?$', '^.*Do you want to save the configuration of the AP\\s?\\?\\s*((\\[yes/no\\]\\??)|(\\[y/n\\]\\??)):?\\s?$', '^(.*)This command will reboot the system.\\s*\\(y\\/n\\)\\?\\s*\\[n\\]\\s?$', '^.*Do you want to enforce secure password standard(\\?)?\\s*\\(yes\\/no\\)(\\s*\\[[yn]\\])?\\:\\s*', '^.*(Enter|Confirm) the password for .*admin', 'Abort( Power On)? Auto Provisioning .*:', '^.*([Uu]sername|(?<!Last )[Ll]ogin): ?$', '^.*[Pp]assword( for )?(\\S+)?: ?$', '^.*Press RETURN to get started.*', 'Press RETURN to get started.*', '^(.*?)Connection.*? closed|disconnect: Broken pipe', '^(.*)Please reset', '^.*(Enter|Confirm) the password for .*admin', '^.*(initial|basic) configuration dialog.*\\s?', '^(.*)Would you like to terminate autoinstall\\? ?\\[yes\\]: $', '__timeout__', '.*The highlighted entry will be (booted|executed) automatically in .*?(\\x1b\\S+)?\\s+', '^.*Are you sure you want to continue\\? \\[y/n\\]\\s?.*$', '^(.*?)\\[((?P<hostname00>([^# \\t\\n\\r\\f\\v\\(\\)]+))|[Ss]witch|[Rr]outer|eWLC).*?\\]\\$\\s?$', '^(.*?)(rommon[\\s\\d]*>|switch:|grub>)\\s*(\\x1b\\S+)?$', '^(.*)\\(config-pki-hexmode\\)#\\s?$', '^(.*?)(\\(unlicensed\\))?(wlc|WLC|Router|RouterRP|Switch|ios|switch|(?P<hostname00>([^# \\t\\n\\r\\f\\v\\(\\)]+)))([0-9])?(\\(standby\\))?(-stby)?(-standby)?(\\(boot\\))?\\(maint-mode\\)#[\\s\\x07]*$', '^(.*?)(\\{\\.\\.\\}|then.else.fi)\\s*>\\s*$', '^(.*?)\\(rules.*?\\)#[\\s\\x07]*$', '^(.*?)\\(syntax.*?\\)#[\\s\\x07]*$', '^(.*?)\\(acm.*?\\)#[\\s\\x07]*$', '^(.*?)\\(tcl.*?\\)#[\\s\\x07]*$', '^(.*)\\[\\S+@guestshell\\s+.*\\][#\\$]\\s?$', '^(.*)\\((?!.*pki-hexmode).*(con|cfg|ipsec-profile|ca-trustpoint|ca-certificate-map|cs-server|ca-profile|gkm-local-server|cloud|host-list|config-gkm-group|gkm-sa-ipsec|gdoi-coop-ks-config|wsma|enforce-rule|DDNS|ca-trustpool|cert-trustpool)\\S*\\)#\\s?$', '^(.*?)(\\(unlicensed\\))?(wlc|WLC|Router|RouterRP|Switch|ios|switch|(?P<hostname00>([^# \\t\\n\\r\\f\\v\\(\\)]+)))([0-9])?(\\(recovery-mode\\))?(\\(rp-rec-mode\\))?(\\(standby\\))?(-stby)?(-standby)?(\\(boot\\))?#[\\s\\x07]*$', '^(?!.*?grub)(.*?)(\\(unlicensed\\))?(wlc|WLC|Router|RouterRP|Switch|ios|switch|(?P<hostname00>([^# \\t\\n\\r\\f\\v\\(\\)]+)))([0-9])?(\\(recovery-mode\\))?(\\(rp-rec-mode\\))?(\\(standby\\))?(-stby)?(-standby)?(\\(boot\\))?(?<! -)>\\s?$', '^.*?([S|s]tandby console disabled|This \\(D\\)RP Node is not ready or active for login \\/configuration.*)', '^.*--\\s?[Mm]ore\\s?--.*$']
             buffer:'\r\n'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "src/unicon/bases/connection.py", line 839, in unicon.bases.connection.Connection.connect
  File "src/unicon/bases/routers/connection_provider.py", line 300, in unicon.bases.routers.connection_provider.BaseSingleRpConnectionProvider.connect
  File "src/unicon/bases/routers/connection_provider.py", line 331, in unicon.bases.routers.connection_provider.BaseSingleRpConnectionProvider.establish_connection
  File "src/unicon/statemachine/statemachine.py", line 742, in unicon.statemachine.statemachine.StateMachine.go_to
unicon.core.errors.StateMachineError: Failed while bringing device to "any" state

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/atestini/Desktop/Automation/testing-for-nac/nac-test/nac_test/pyats_core/broker/connection_broker.py", line 331, in _create_connection
    await loop.run_in_executor(
  File "/Users/atestini/.pyenv/versions/3.12.9/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/atestini/Desktop/Automation/testing-for-nac/nac-test/nac_test/pyats_core/broker/connection_broker.py", line 333, in <lambda>
    lambda: device.connect(log_stdout=False, logfile=str(logfile_path)),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/pyats/connections/manager.py", line 511, in pyats.connections.manager.ConnectionManager.connect
  File "src/unicon/bases/connection.py", line 850, in unicon.bases.connection.Connection.connect
unicon.core.errors.ConnectionError: failed to connect to iosxe-r1
Failed while bringing device to "any" state
2026-02-16T18:47:26: %NAC_TEST-ERROR: Error communicating with broker: Broker error: Unknown broker error
2026-02-16T18:47:26: %NAC_TEST-ERROR: Failed to ensure connection to iosxe-r1: Broker error: Unknown broker error
2026-02-16T18:47:26: %SCRIPT-ERROR: Failed to connect to device iosxe-r1: Failed to connect to device: iosxe-r1
2026-02-16 18:47:26.616 [PID:78350] [76947] [ID:3] FAILED operational.verify_ospf_neighbor_state in 85.9 seconds
------------------------------------------------------------------------------
Verify All OSPF Neighbors Are in FULL State (iosxe-r1)                 | FAILED |
------------------------------------------------------------------------------

================================================================================
3 tests, 1 passed, 2 failed, 0 skipped.
================================================================================

PyATS Output Files:
================================================================================
Results JSON:    /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results/pyats_results/api/results.json
Report:          /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results/pyats_results/api/tmpfwmpcd2h_api_job.report
Archive:         /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results/nac_test_job_api_20260216_184423_800.zip

Archive:         /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results/nac_test_job_d2d_20260216_184729_004.zip

Total testing: 2 minutes 53.47 seconds
Elapsed time:  3 minutes 5.32 seconds

Generating reports from 2 archive(s)...
Total report generation time: 0.10 seconds
Found API archive: nac_test_job_api_20260216_184423_800.zip
Found D2D archive: nac_test_job_d2d_20260216_184729_004.zip

HTML Reports Generated:
================================================================================
Combined Summary: /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results/pyats_results/combined_summary.html
API Reports:   /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results/pyats_results/api/html_reports
D2D Reports:   /Users/atestini/Desktop/Automation/testing-for-nac/nac-sdwan-terraform/nac-test-results/pyats_results/d2d/html_reports
```

---

oboehmer:
## Description

Addresses three related issues in the `ConnectionBroker` / `BrokerClient` / `SSHTestBase` stack that together made D2D test failures difficult to diagnose and the broker path unreliable.

**Noisy connection error output (#539):** When a device was unreachable, users saw ~100 lines of Unicon internal state machine tracebacks, a generic "Unknown broker error" message, and the same failure logged 4 times across different call layers. This was introduced by #467, which reordered the broker/testbed priority so the broker path started being exercised — activating `exc_info=True` in `_create_connection` for the first time. Fixed by: dropping `exc_info=True`, propagating the real exception message through the broker response (was returning `{"status": "error", "result": false}` with no `"error"` key, causing the fallback to "Unknown broker error"), and reducing intermediate log levels to `debug` so the final user-facing ERROR appears exactly once.

**Broker failure-mode test coverage (#486):** The existing broker tests only covered the happy path. This PR adds integration tests that exercise the broker against mocked pyATS devices at the `device.connect()` boundary, covering: broker unreachable, device unreachable, disconnect-on-error, and full shutdown. Using a real Unix socket and real async protocol means these tests catch bugs that pure unit tests cannot reach.

**Deadlock in `BrokerClient.connect()` on failure (discovered via #486):** Writing the failure-mode tests uncovered a deadlock: when the broker returned a connection error, `BrokerClient.connect()` did not release internal state correctly, leaving the client permanently stuck. The `test_ensure_connection_returns_false_when_broker_not_running` test would have hung indefinitely without the fix.

**Broker socket validation and graceful fallback (#485):** Previously, the broker path was taken whenever `NAC_TEST_BROKER_SOCKET` was set (presence-only). This PR validates that the env var points to a real Unix socket via `Path.is_socket()`. If it’s set but invalid (missing path, regular file, directory), a warning is logged and we fall back to direct testbed connection.

#### Unit tests (`tests/unit/pyats_core/test_connection_broker.py`)
Covers broker-side request processing and health checks, including:
- `_is_connection_healthy()` edge cases (attribute access errors / missing attrs)
- `_process_request()` missing-field error messages are specific (e.g., mentions `hostname` / `cmd`)

#### Unit tests (`tests/unit/test_ssh_base_test_validation.py`)
Covers broker socket env handling:
- Use broker when `NAC_TEST_BROKER_SOCKET` points to a valid Unix socket
- Warn + fall back when it points to missing path / regular file / directory
- Raise when broker socket is invalid and no testbed device exists

#### Integration tests (`tests/integration/test_connection_broker_failure_modes.py`)
Covers socket protocol + client/broker interaction, including:
- Broker unavailable: missing socket, stale socket, socket path is dir/file
- Device connect failure: error message preserved (not “Unknown broker error”)
- Socket lifecycle: unlink socket mid-run (existing connection works; new client fails)
- Connection health: reconnect when cached connection unhealthy
- Concurrency: concurrent connect requests dedupe to one device.connect (single-process + multi-process)
- Cleanup: disconnect cleanup even when device.disconnect raises; shutdown disconnects devices + removes socket
- Protocol/request hardening:
  - Unknown/missing command errors
  - Missing required parameters (parametrized)
  - Invalid protocol closes connection (parametrized): truncated frame, non-JSON, bad UTF-8, zero-length, absurd length prefix
  - Valid JSON but wrong type returns framed JSON error response
  - Sequential requests over one connection: handled error does not poison the connection; invalid protocol closes it

### Timeout enhancements
Follow-up work tracked in **#735** (BrokerClient timeouts + detailed errors).

### #486 scenarios not covered in this PR (and why)
Some requested scenarios require product-level behavior decisions and/or timeouts to be non-flaky:
- **Broker crash mid-run** (undefined desired behavior: abort vs fallback vs restart) → **#736**
- **Command execution / socket IO timeout behavior** (needs product timeout semantics, not just `wait_for` in tests) → **#735**

To ensure these remain tracked after closing #486, I created: **#737** “test: cover remaining ConnectionBroker failure modes (crash + timeouts)”.

## Closes

- Fixes #539
- Fixes #486
- Fixes #485

## Related Issue(s)

- #735 
- #736 
- #737 

## Type of Change

- [ ] Documentation update

## Test Framework Affected

- [x] PyATS
- [ ] Robot Framework
- [ ] Both

- [ ] ACI (APIC)
- [ ] Meraki (Cloud-managed)
- [x] NX-OS (Nexus Direct-to-Device)
- [x] IOS-XE (Direct-to-Device)
- [x] IOS-XR (Direct-to-Device)
- [ ] Hyperfabric
- [ ] All architectures
- [x] N/A (architecture-agnostic)

## Platform Tested

- [x] macOS (version tested: 15.x)

## Key Changes

- `connection_broker.py`: drop `exc_info=True` from `_create_connection`, raise on failure instead of returning `None`, propagate the real error message in the broker response
- `broker_client.py`: fix deadlock in `connect()` on connection failure; downgrade intermediate error logs to `debug` to eliminate duplicate console output
- `ssh_base_test.py`: validate `NAC_TEST_BROKER_SOCKET` contents with `Path.is_socket()`
- Tests: see above

## Testing Done

### Test Commands Used

```bash
pytest tests/unit/ -q
pre-commit run -a
```

## Checklist

## Additional Notes

The `BrokerClient.connect()` deadlock was not directly reported — it was found while writing the failure-mode test coverage requested in #486. That test (`test_ensure_connection_returns_false_when_broker_not_running`) would have hung indefinitely without the fix.


aitestino (changes requested):
Hey @oboehmer, thank you for the PR -- this is solid, important work. Fixing a real deadlock in `BrokerClient.connect()`, eliminating the confusing ~100-line Unicon traceback noise, and hardening the broker socket validation are all things that needed to happen. The `_teardown_connection()` extraction is the textbook correct fix for the lock re-entry deadlock -- the docstring explaining why `wait_closed()` is skipped on the connect-failure path shows real care for the next person reading this code. The `Path.is_socket()` validation with fallback-and-warning is exactly the right pattern (warn but don't block). And the 1,300+ lines of new failure-mode tests -- especially the multi-process concurrency test that spawns 5 real subprocesses against a live broker socket -- are genuinely impressive. The protocol-level tests (truncated frames, bad UTF-8, absurd length prefix) are the kind of coverage that prevents real production surprises.

I did a couple of reviews on this, and I think we need some changes before merge. The core design is sound -- the deadlock fix is correct, the error propagation threading is right, and the integration test strategy (real socket, mock at the `device.connect()` boundary) is well-chosen. These are mostly structural and consistency items.

1. **Server teardown boilerplate repeated 19 times** -- The `assert broker.server is not None / broker.server.close() / await broker.server.wait_closed()` block appears in every single test method in the integration file. An `asynccontextmanager` helper (same pattern as `_patch_executor` already in the file) would collapse all 19 copies into one definition. Something like:

   ```python
   @asynccontextmanager
   async def _running_broker(broker: ConnectionBroker) -> AsyncGenerator[ConnectionBroker, None]:
       await broker._start_socket_server()
       try:
           yield broker
       finally:
           assert broker.server is not None
           broker.server.close()
           await broker.server.wait_closed()
   ```

   This is the highest-ROI change -- 57 duplicated lines collapsed to a single definition, and any future teardown change happens in one place.

2. **`_create_connection` has an inconsistent error contract** -- The PR description says it now "raises instead of returning None," which is true for the `device.connect()` failure path. But the two guard clauses at the top (no testbed loaded, hostname not in testbed) still silently return `None`, and the return type annotation still says `Any | None`. This means the method has two different failure signaling mechanisms in one function. Since the callers already wrap this in try/except (via `_ensure_connection`), converting those two guards to also raise with specific messages (e.g., `raise ConnectionError(f"No testbed loaded for {hostname}")`) would complete the contract change and let the annotation tighten to `-> Any`.

3. **`socket_dir` fixture defined twice in the same PR** -- The fixture is added to `tests/unit/conftest.py` (properly typed, `Generator[Path, None, None]`, top-level import) and then re-defined locally in the integration test file (typed as `-> Any`, inline `import tempfile`). Since both `tests/unit/` and `tests/integration/` need it, the right home is `tests/conftest.py` at the root level, with both copies deleted.

4. **`test_connection_broker.py` placed one level too high** -- The source under test lives at `nac_test/pyats_core/broker/`. The test structure everywhere else mirrors the source path (`common/ -> tests/unit/pyats_core/common/`, `execution/ -> tests/unit/pyats_core/execution/`). This file should be at `tests/unit/pyats_core/broker/test_connection_broker.py` with an `__init__.py`.

5. **Hostname injection in `_create_connection` is fragile** -- The `if hostname not in msg` check is a case-sensitive substring match, and the log message drops the exception type entirely. If Unicon raises `RuntimeError: connection reset by peer`, the log is just `"hostname: connection reset by peer"` -- no exception type, less useful than it could be. A fixed-format message like `logger.error(f"Failed to connect to {hostname}: {e}")` is more robust and keeps the diagnostic quality high.

6. **Missing `pytestmark = pytest.mark.integration`** -- Every other file in `tests/integration/` declares this marker. Without it, `pytest -m integration` won't pick up these tests.

7. **`connect()` in BrokerClient still logs at ERROR before raising** -- The PR correctly downgraded `_send_request` and `ensure_connection` from `logger.error` to `logger.debug` to eliminate duplicate output. But `connect()` at line 75 still calls `logger.error(...)` before raising `ConnectionError`, and the caller (`_async_setup`) catches and logs again. Applying the same `debug` downgrade here would be consistent with the PR's own stated goal.

8. **No max frame size enforcement on the broker protocol** -- `_handle_client` reads a 4-byte length prefix and calls `reader.readexactly(message_length)` with no upper bound. A 4-byte unsigned integer can represent ~4 GB. Adding a `MAX_BROKER_MESSAGE_BYTES` constant (say, 10 MB) and rejecting oversized frames before `readexactly()` would prevent a buggy client from exhausting broker memory. I see #735 is tracking timeout work -- this could bundle naturally there.

9. **Five tautological unit tests in `test_connection_broker.py`** -- `test_execute_command_sends_correct_request`, `test_ping_returns_true_on_pong`, `test_disconnect_device_sends_correct_request`, `test_connect_success`, and `test_disconnect_calls_disconnect_device` all assert that mocks return mocked values. For example, `test_ping_returns_true_on_pong` mocks `_send_request` to return `{"result": "pong"}` and then asserts `ping()` returns `True` -- which is just testing that `"pong" == "pong"`. The integration tests already exercise these paths through the real socket. The companion test `test_ping_returns_false_on_failure` (which tests exception-swallowing behavior) IS valuable and should stay.

10. **Two tests in the integration file are actually unit tests** -- `TestConnectionHealthCheckFailure` calls `broker._get_connection()` directly with no socket involved, and `TestDisconnectCleanup` calls `broker._disconnect_device()` directly. Neither exercises any socket/IPC path. They're valuable tests -- they just belong in the unit file.

11. **`assert` statements used for runtime invariants** -- `broker_client.py:115-116` uses `assert self.writer is not None` and `assert self.reader is not None`. These are stripped when Python runs with `-O`. Should be `if ... raise RuntimeError(...)` guards instead.

I'd consider items 1-6 as things to address before merge -- they're all straightforward and most are mechanical. Items 7-11 are lower priority and could be follow-ups if you'd rather not hold this up further.

What do you think?

oboehmer:
Hey @aitestino, thanks for the thorough review — all 11 items addressed. Each has its own commit so you can review them individually.

**Items 1–6 (pre-merge):**

1. **Teardown boilerplate** → f4a85f1 — Converted the 4 remaining inline teardown blocks to the existing `_run_broker` helper (the file already had this pattern; the actual count was 4 inline instances, not 19, since `_run_broker` was already used for most tests). Net reduction is ~20 lines of duplicated teardown.

2. **`_create_connection` error contract** → ff76be4 — Both guard clauses now raise (`RuntimeError` for no testbed, `ConnectionError` for unknown hostname) instead of returning `None`. Return annotation tightened from `Any | None` to `-> Any`. Callers in `_ensure_connection` already use try/except so this is seamless.

3. **`socket_dir` fixture duplication** → 715d9bf — Moved to `tests/conftest.py` (root level), removed from both `tests/unit/conftest.py` and the inline definition in the integration file. Proper `Generator[Path, None, None]` typing. This introduced a shared root conftest — opened #764 to audit and consolidate other duplicate fixtures across the test suite into it.

4. **Test file placement** → 693f2fc — Moved to `tests/unit/pyats_core/broker/test_connection_broker.py` with `__init__.py`, mirroring the source path.

5. **Hostname injection in `_create_connection`** → 83c0869 + d2db443 — Kept the `if hostname not in str(e)` guard because pyATS exceptions frequently embed the hostname (e.g. "failed to connect to iosxe-r1"), and always prepending "Failed to connect to iosxe-r1:" produces redundant output. Added a comment explaining the rationale. The format is now `f"{type(e).__name__}: {e}"` (with hostname prefix only when absent), so the exception type is always included. Both paths have dedicated unit tests.

6. **Missing `pytestmark`** → 2eda128 — Added `pytestmark = pytest.mark.integration`.

**Items 7–11 (also addressed in this PR):**

7. **`connect()` ERROR→DEBUG** → b11053d — Downgraded to `logger.debug`, consistent with the other methods.

8. **`MAX_BROKER_MESSAGE_BYTES`** → 0e0cbae — Implemented here rather than deferring to #735. 10 MB constant defined in `nac_test/pyats_core/constants.py`, enforced in `_handle_client` before `readexactly()`. Includes an integration test that sends an oversized length prefix and verifies the connection is closed.

9. **Tautological unit tests** → 0f0ea46 — Removed 5 mock-returns-mock tests. Kept `test_ping_returns_false_on_failure` (tests exception-swallowing behavior).

10. **Integration→unit moves** → 4a6d1f1 — Moved `TestConnectionHealthCheckFailure` and `TestDisconnectCleanup` to the unit file since they don't exercise any socket path.

11. **`assert` → `RuntimeError`** → af88e7c — Replaced `assert self.writer/reader is not None` with `if ... raise RuntimeError(...)` guards.

oboehmer:
merging, happy to follow-up any comments you might still have on my latest changes addressing your comments.

PR #749, opened 2026-04-13
Chore/relax aiofiles pin

oboehmer:
Relax dependency for `aiofiles` to `>=23.2.1` in order to avoid conflicts for concurrent nac-test and radkit-client installation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
         exclude: 'tests/fixtures/|tests/integration/fixtures/|tests/e2e/fixtures/'
         additional_dependencies:
           - types-PyYAML>=6.0.12.20250822
-          - types-aiofiles>=24.1.0.20250822
+          - types-aiofiles>=23.2.0.20240623
           - types-markdown>=3.8.0.20250809
           - types-requests>=2.32.0.20250809
   - repo: https://github.com/PyCQA/bandit
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "genie>=25.5 ; sys_platform != 'win32'",
     "psutil>=7.0.0",
     "httpx~=0.28",
-    "aiofiles>=24.1",
+    "aiofiles>=23.2.1",     # radit-client still uses <24.0, so relax to avoid conflicts
     "colorama>=0.4.6",
     "markdown>=3.8.2",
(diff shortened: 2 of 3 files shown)


aitestino (approved):
Hey @oboehmer, thank you for the PR -- relaxing the `aiofiles` floor to unblock `radkit-client` co-installation is the right call. The motivation is clear, the approach is minimal, and the aiofiles API surface used in nac-test (`aiofiles.open()` with async context managers and async iteration) has been rock-solid across every version from 23.x through 25.x, so there's no regression risk from the relaxation itself.

I did a couple of reviews on this, and I think we need a few small tweaks before merge. The core change is sound -- these are all polish items.

1. **Typo in the inline comment** -- `radit-client` should be `radkit-client`. This comment is the only documentation for why the floor was lowered, so it needs to be searchable and accurate. One-character fix.

2. **`.pre-commit-config.yaml` not updated** -- The mypy hook at line 34 still pins `types-aiofiles>=24.1.0.20250822`, while `pyproject.toml` was relaxed to `>=23.2.0.20240623`. These should match for consistency. Quick find-and-replace.

3. **Inline comment spacing** -- The new comment uses 5 spaces before `#`, but the existing convention in the file is 2 spaces (see the `jsonpath-ng` and `robotframework-pabot` lines). Minor, but worth normalizing while you're in there.

I'd consider items 1-2 as things to address before merge -- they're both one-liner fixes. Item 3 is a nitpick that could ride along with the typo fix.

A couple of observations that don't need action but are worth noting:

- The `types-aiofiles` relaxation makes sense for consistency (stubs tracking the runtime floor), though it's worth calling out that it's not strictly required by the radkit-client conflict since dev dependencies aren't involved in the co-installation scenario. Might be worth a brief note in the PR description so future readers understand the intent.

- The `uv.lock` diff also shows the `httpx` specifier catching up from `>=0.28` to `~=0.28` and the version bumping from `2.0.0b2` to `2.0.0b3` -- both of these are just the lock file re-syncing with what's already on `main`, not changes introduced by this PR. No action needed, just noting for anyone reviewing the diff who might wonder.

What do you think?

_P.S. -- This comment was drafted using voice-to-text via Claude Code. If the tone comes across as overly direct or terse, please know that's just how it tends to phrase things. No offense or criticism is intended -- this is purely an objective technical review of the PR. Thanks for understanding!_

oboehmer:
Thanks, Andrea! Addressed all comments.

Issue #663, opened 2026-03-19
[bug]  Genie BGP Parser Returns Inconsistent VRF Keys Causing False Test Failures


koval4ukav:
### Bug Description

Data model driven D2D test`verify_bgp_peers` always fails with **"BGP Neighbor Not Found"** for any
BGP peer that lives in a service VRF (VPN ID 1–512), even when the BGP session is fully
established on the device. The failure is a false negative produced by an incorrect VRF key
lookup in the Genie-parsed output — the device and its BGP sessions are healthy.

### Test Framework

PyATS (Python-based tests)

### Test Type (PyATS Only)

D2D/SSH Tests (direct-to-device - tests/d2d/)

### Network as Code (NaC) Architecture

Catalyst SD-WAN (SDWAN Manager / vManage)

### Operating System

Linux

### OS Version

Ubuntu 22.04

### Installation Method

uv (uv pip install nac-test)

### System Information Output

```text
=== System Info ===
NAME="Ubuntu"
VERSION="22.04.5 LTS (Jammy Jellyfish)"

$ uv run nac-test --version
nac-test, version 2.0.0a1

Python 3.12.11
```

### Environment Variables Output

```bash
SDWAN_URL=*** SET ***
SDWAN_USERNAME=*** SET ***
SDWAN_PASSWORD=*** SET ***
NAC_TEST_DEBUG=true
```

### Reproduction Steps

1. Configure a Catalyst SD-WAN edge (C8000v or similar IOS-XE device) with BGP in a service
   VPN (e.g. VPN ID 10, neighbor `10.10.1.10`, remote-as `64555`).
2. Define the BGP neighbor in the NaC data model under a service VPN template (UX1) or a
   service profile BGP feature (UX2).
3. Run the `verify_bgp_peers.py` D2D test against the device.
4. Observe: test fails with "BGP Neighbor Not Found" even though
   `show ip bgp all summary` on the device shows the session as established.

### Command Used

```bash
uv run nac-test -d data -d defaults -t ./tests  -o ./res-260 -f tests/jinja_filters --verbose --loglevel DEBUG
```

### Data Model Snippet (if relevant)

```yaml
# UX2 (configuration group) — DC hub with BGP in service VPN 10
sdwan:
  configuration_groups:
    - name: dc_hub
      service_profile: service_lan_bgp
      device_tags:
        - name: dc_primary
          features:
            - lan_vpn10_bgp_primary
            - bgp_feature_vpn10_primary

  feature_profiles:
[...]

PR #835, opened 2026-05-19
Consolidation of Genie parsing logic to Connection Broker

seojumper:
## Description

PR consolidates genir parsing logic to ConnectionBroker with storage of parsed output retained in CommandCache.

## Related Issue(s)

This is related to #663 where previously parsing of previously fetched command output that already existed in cache was not working correctly based on discovered logic in Genie Parsers. 

## Type of Change

- [X] Refactoring / Technical debt (internal improvements with no user-facing changes)
- [ ] Documentation update

## Test Framework Affected

- [X] PyATS
- [ ] Robot Framework
- [ ] Both

- [ ] ACI (APIC)
- [ ] Meraki (Cloud-managed)
- [X] IOS-XE (Direct-to-Device)
- [ ] Hyperfabric
- [ ] All architectures

## Platform Tested

- [ ] macOS (version tested: )
- [X] Linux (distro/version tested: Unbuntu 24.04)

## Testing Done

- [X] Unit tests added/updated
- [X] Integration tests performed
- [X] Manual testing performed:
  - [X] PyATS tests executed successfully
  - [X] Robot Framework tests executed successfully
  - [X] D2D/SSH tests executed successfully (if applicable)
  - [X] HTML reports generated correctly
- [X] All existing tests pass (`pytest` / `pre-commit run -a`)

### Test Commands Used

```bash
uv run pytest -n auto --dist loadscope tests/
```

## Checklist

- [X] Code follows project style guidelines (`pre-commit run -a` passes)
- [X] Self-review of code completed
- [X] Code is commented where necessary (especially complex logic)
- [X] No new warnings introduced

## Additional Notes

This change is currently causing tests to directly call 'parse_output' instead of 'execute_command' > 'parse_output' which does impact visual of test results as raw command output is no longer present.  Default genie api's do not return command output from a 'parse()' call.  This could be changed with providing the 'return_raw=True' optional arg to the 'parse_output' fn.

### DIFF
--- a/nac_test/pyats_core/broker/broker_client.py
+++ b/nac_test/pyats_core/broker/broker_client.py
@@ -126,6 +126,7 @@ class BrokerClient:
 
             # Read response data
             response_data = await self.reader.readexactly(response_length)
+            logger.debug(f"Received response of length: {response_length} bytes")
             response = json.loads(response_data.decode("utf-8"))
 
             # Check for errors
@@ -164,6 +165,37 @@ class BrokerClient:
         logger.debug(f"Command output length: {len(result)} characters")
 
         return result  # type: ignore[no-any-return]
+
+    async def parse_command(
+        self, hostname: str, command: str, return_raw: bool = False
+    ) -> dict[str, Any]:
+        """Parse command output on device through broker.
+
+        Args:
+            hostname: Device hostname
+            command: Command to parse
+            return_raw: Whether to return raw output along with parsed result
+
+        Returns:
+            Tuple containing parsed output dictionary and optionally raw output
+
+        Raises:
+            ConnectionError: If broker communication fails
+        """
+        logger.debug(
+            f"Parsing command on {hostname}: {command} (return_raw={return_raw})"
+        )
+
+        response = await self._send_request(
+            {
+                "command": "parse",
+                "hostname": hostname,
+                "cmd": command,
+                "return_raw": return_raw,
+            }
+        )
+
+        return response.get("result", {})  # type: ignore[no-any-return]
 
     async def ensure_connection(self, hostname: str) -> bool:
         """Ensure device is connected through broker.
@@ -257,6 +289,23 @@ class BrokerCommandExecutor:
             Command output
         """
         return await self.broker_client.execute_command(self.hostname, command)
+
+    async def parse(
+        self, command: str, return_raw: bool = False
+    ) -> tuple[dict[str, Any], str | None]:
+        """Parse command output on device via broker.
+
+        Args:
+            command: Command to parse
+            return_raw: Whether to return raw output along with parsed result
+
+        Returns:
+            Dictionary containing parsed output and optionally raw output
+        """
+        result = await self.broker_client.parse_command(
+            self.hostname, command, return_raw
+        )
+        return result.get("parsed", {}), result.get("raw", None)
 
     async def connect(self) -> None:
         """Ensure device connection via broker."""
--- a/nac_test/pyats_core/broker/connection_broker.py
+++ b/nac_test/pyats_core/broker/connection_broker.py
@@ -199,6 +199,19 @@ class ConnectionBroker:
 
                 result = await self._execute_command(hostname, cmd_string)
                 return {"status": "success", "result": result}
+            elif command == "parse":
+                hostname = message.get("hostname")
+                return_raw = message.get("return_raw", False)
+                cmd = message.get("cmd")
+
+                if not hostname or not cmd:
+                    return {
+                        "status": "error",
+                        "error": "Missing hostname or cmd parameter",
+                    }
+
+                result = await self._parse_command(hostname, cmd, return_raw)
+                return {"status": "success", "result": result}
 
             elif command == "connect":
                 hostname = message.get("hostname")
@@ -253,9 +266,10 @@ class ConnectionBroker:
         # Check cache first
         cached_output = cache.get(cmd)
         if cached_output is not None:
-            self.stats_command_cache_hits += 1
-            logger.debug(f"Broker cache hit for '{cmd}' on {hostname}")
-            return cached_output
+            if cached_output.output is not None:
+                self.stats_command_cache_hits += 1
+                logger.debug(f"Broker cache hit for '{cmd}' on {hostname}")
+                return cached_output.output
 
         # Command not in cache, need to execute
         self.stats_command_cache_misses += 1
@@ -271,7 +285,7 @@ class ConnectionBroker:
             output_str = str(output)
 
             # Cache the output for future requests
-            cache.set(cmd, output_str)
+            cache.set(cmd, output=output_str)
             logger.info(
                 f"Cached command output for '{cmd}' on {hostname} ({len(output_str)} chars)"
             )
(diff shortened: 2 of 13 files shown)


oboehmer:
Thanks for the fix, @seojumper ! Can we also fix this without having to adjust our already generated test cases?

oboehmer:
Thanks for digging into this, Jeremy. I now reviewed the underlying issue, and I'd prefer a lighter fix that addresses the root cause directly rather than restructuring the parse flow into the broker.

**Root cause:** Some genie parsers call `self.device.execute()` internally for supplementary commands (e.g., `show vrf`, `show run | sec address-family`) to resolve VRF names. In broker mode, the testbed device in the test process isn't connected — those calls fail silently, so VRF resolution degrades and everything collapses to `"default"`.

**Proposed fix:** Patch `testbed_device.execute` to route through the broker, so Genie's internal calls work transparently. No new RPC, no `CacheEntry` restructuring, no broker-side parse logic needed.

**Sketch:**

```python
from functools import partial

# In _async_setup(), after broker connection is established:
if self.testbed_device:
    self._patch_device_execute_for_broker()

def _patch_device_execute_for_broker(self):
    """Patch testbed_device.execute so Genie's internal calls route through the broker."""
    broker_client = self.broker_client
    hostname = self.hostname
    loop = get_or_create_event_loop()

    def broker_execute(cmd, *args, **kwargs):
        """Sync execute that routes through connection broker.

        Must be called from a worker thread (via run_in_executor),
        not from the event loop thread, to avoid deadlock.
        """
        future = asyncio.run_coroutine_threadsafe(
            broker_client.execute_command(hostname, cmd), loop
        )
        return future.result(timeout=120)

    self.testbed_device.execute = broker_execute
```

`parse_output` becomes async and runs Genie in a worker thread (required because the patched execute blocks waiting for the broker response — safe from a thread, deadlocks on the event loop thread):

```python
async def parse_output(
    self, command: str, output: str | None = None
) -> dict[str, Any] | None:
    """Parse command output using Genie parser.

    Runs in a worker thread so Genie's supplementary device.execute()
    calls can safely route through the broker without deadlocking.
    """
    if not self.testbed_device:
        return None
    try:
        loop = get_or_create_event_loop()
        if output is not None:
            result = await loop.run_in_executor(
                None, partial(self.testbed_device.parse, command, output=output)
            )
        else:
            result = await loop.run_in_executor(
                None, partial(self.testbed_device.parse, command)
            )
        return dict(result) if result is not None else None
    except Exception as e:
        self.logger.warning(f"Genie parser failed for '{command}': {e}")
        return None
```

**Impact on test templates — one keyword added:**

```python
# Before:
parsed_output = self.parse_output(command, output=output)

# After:
parsed_output = await self.parse_output(command, output=output)
```

**Why I prefer this:**
- Fixes the root cause (Genie can't execute supplementary commands) rather than working around it
- No new broker RPC (`parse` command), no `CacheEntry` class, no `_parse_command` method
- Raw command output stays in the test flow — available for reporting without `return_raw` plumbing
- Broker-side `CommandCache` already deduplicates repeated executions from supplementary calls
- Smaller diff, less surface area for bugs
- only minimal changes required for the test cases

Happy to discuss further — the key insight is that Genie just needs a working `device.execute()` and it handles everything else.

can you please try out this approach, @seojumper ?

oboehmer:
I pushed the alternative approach to https://github.com/netascode/nac-test/compare/663-genie-device-execute-support, could you test this in your environment? I also added a test which reproduces the bug, so we can also catch regressions here in our pipeline. @aitestino @seojumper

seojumper:
> I pushed the alternative approach to https://github.com/netascode/nac-test/compare/663-genie-device-execute-support, could you test this in your environment? I also added a test which reproduces the bug, so we can also catch regressions here in our pipeline. @aitestino @seojumper

yeah i can try that out tonight/tomorrow, thanks.

aitestino:
Hey @seojumper, thank you for the PR — the intent here is correct, and moving the Genie parsing logic into a more centralized path is definitely the right direction for solving #663.

Per our internal conversation but recapping here for business continuity, it looks like Oliver (@oboehmer) has created his own branch (`663-genie-device-execute-support`) with an alternative implementation. I haven't seen a PR for it yet, but I've gone through both implementations in detail. After comparing the two approaches, I think Oliver's is the stronger foundation here for a few key reasons:

**Why Oliver's approach fits the codebase better:**

1. **Cache coherence** — The implementation in this PR introduces two writers to the same `CommandCache` (`_execute_command` and `_parse_command`) that can overwrite each other's entries. When `_parse_command` is called with `return_raw=False`, it stores `output=None`. A subsequent `_execute_command` for the same command sees `output is None`, treats it as a miss, re-executes on the device, and then writes a new `CacheEntry` that destroys the cached `parsed_output`. The reverse scenario also applies. This is the cache-swapping concern you flagged yourself, and it's a real bug, not just theoretical.

2. **Genie supplementary commands** — The core issue (#663) is that Genie parsers internally call `device.execute()` for supplementary data. The approach in this PR moves `connection.parse()` into the broker, but Genie's supplementary `execute()` calls during that parse still go through the broker's direct connection — which works, but only because the broker owns the connection.
   - Oliver's approach solves this transparently at the right layer: ALL `device.execute()` calls (explicit test calls AND Genie's internal ones) route through the same path.

3. **Backward compatibility** — The change in this PR removes the `output=` parameter from `parse_output()` and changes the return type from `dict | None` to `tuple[dict | None, str | None]`. This breaks downstream consumers in `nac-sdwan-terraform` (`verify_bgp_peers.py`, `verify_ospf_neighbor_state.py`).
   - Oliver's approach preserves the parameter signature.

4. **SRP** — From a best practices standpoint, I think ideally we want to keep the broker as a thin command transport layer. Adding parsing responsibility to the broker creates coupling between the connection management layer and Genie's parser internals.

We also discussed some concerns surrounding monkey patching annnnd I hear you, and I share the general instinct that monkey patching creates behavior that's harder to reason about. It's a legitimate tradeoff. In this specific case though, it's the only approach that transparently intercepts Genie's internal `execute()` calls without modifying pyATS source (which we can't do).

Oliver's patching is well-isolated to one method (`_patch_device_execute_for_broker`), well-documented with clear docstrings explaining WHY each patch exists, and the integration tests would immediately surface any breakage if pyATS changes their dispatch mechanism. Given the alternative is a design-level cache coherence problem, I think the monkey-patch tradeoff is the pragmatic call here.

---

@oboehmer — Looking at the work in your branch, here are my findings/thoughts:

**Things that likely need adjustment:**

1. **`get_cache_stats()` is not thread-safe** — You correctly added `threading.Lock` to `get()`, `set()`, and `clear()`, but `get_cache_stats()` iterates `self.cache.values()` without the lock. Under concurrent access from worker threads this can raise `RuntimeError: dictionary changed size during iteration`. One-line fix: wrap it in `with self._lock`.

2. **Broker-without-testbed regression** — `_create_execute_command_method` captures `device = test_instance.testbed_device` and always calls `device.execute`. If the broker is active but no testbed file is loaded, `device` is `None` and you get an `AttributeError`. The old code dispatched correctly on both paths. Either guard with an explicit assert or restore the fallback.

3. **Dead `connection` parameter** — `_create_execute_command_method(self, connection, command_cache)` accepts `connection` but never uses it. The docstring says "kept for API compatibility" but it's a private method with one call site. Remove it or use it as the fallback for the `device is None` case (which would also fix issue #2).

4. **`test_broker_genie_parse_2.py` TITLE copy-paste** — Both files say `"(1)"` in the TITLE constant. File 2 should say `"(2)"`.

5. **`DEVICE_EXECUTE_TIMEOUT` missing from `__all__`** — Every other constant in that file is exported. Add it for consistency.

6. **`# type: ignore[assignment]` on hostname/command_cache** — Replace with explicit `assert self.hostname is not None` / `assert self.command_cache is not None` guards. That way the precondition is enforced in code, not suppressed.

7. **Unit test gap for the shim/connected-flag patching** — The existing unit tests cover `broker_execute` routing well, but there's no test verifying that `testbed_device.connected = True`, `connectionmgr.is_connected`, and `device.cli = _BrokerCliShim()` are set correctly. These are the exact attributes that make Genie's parser dispatch work — worth a focused test.

8. **DRY on genie parse fixtures** — `test_broker_genie_parse_1.py` and `_2.py` are 216-line clones differing only in class name. I'd suggest a shared base module with the `verify_item` logic, and two thin subclass files. Same feedback applies to the pre-existing `test_broker_pooling_1/2/3.py` if you want to tackle it in the same pass.

Rest looks good to me.

Once those items are addressed, I think this is ready to go. Would you mind opening a PR for the branch so we can track it properly?

What do you both think?

oboehmer:
Thanks for your comprehensive review of the situation and the two approaches, @aitestino ! I share the same concern about putting more logic into the broker (and we couldn't stop here if people would leverage other genie applications like `learn()` in their tests), so we would end up keeping enhancing the RPC. Keeping it slim and focused on device connection and command execution is IMHO the right call and triggered me trying the alternative approach in my branch (wanted to see if it works before commenting).

To your concern about patching `device.execute()`: I have been following the same design approach in `radkit_genie` for a few years (since its inception), and ran it with the pyATS architects back then (as I was also not sure if this is solid). They voiced no concern about this, and I have not seen any issues with this approach in LazyMaestro/RADKit over the years it has been in production and heavily used. The pyATS API has proven to be very stable, and if it ever changes, we will see it early as we have e2e tests with real pyATS testbed devices exercising all relevant call paths.

I will raise a PR soon, taking your comments into account.

oboehmer:
I raised PR #863

PR #847, opened 2026-05-27
Add sdwan (20.18+) jwt token based auth support for pyats nac-test testcases

Anush-Mangaluru:
## Description

Introduces the `CredentialSet` dataclass and multi-credential-set support to `CONTROLLER_REGISTRY`, replacing the old `CREDENTIAL_PATTERNS` dictionary. Each controller can now define multiple ordered credential combinations (e.g., SDWAN supports both API token and username/password). The first fully-satisfied set wins, and the winning `CredentialSet` is exposed via `get_matched_credential_set()` for downstream auth adapters in nac-test-pyats-common.

Also simplifies `_find_credential_sets()` — removes per-variable present/missing tracking (`CredentialSetStatus` TypedDict) in favor of a simple boolean `has_any_var` check. Incomplete credential error messages now show "incomplete credentials" with all accepted credential sets listed, rather than reporting specific missing variables (the code to keep the previous behavior is more complex, with little gain, really [edited oboehmer]).

## Closes

- https://github.com/netascode/nac-test/issues/854 (feature request)

## Related Issue(s)

- Related to nac-test-pyats-common issue : https://github.com/netascode/nac-test-pyats-common/issues/34
- Parent feature request - - Internal NAC-SDWAN issue #1110

## Type of Change

- [x] Documentation update

## Test Framework Affected

- [ ] PyATS
- [ ] Robot Framework
- [x] Both

- [ ] ACI (APIC)
- [x] Catalyst SD-WAN (SDWAN Manager / vManage)
- [ ] Meraki (Cloud-managed)
- [ ] Hyperfabric
- [x] All architectures

## Platform Tested

- [x] macOS (version tested: macOS with Python 3.13.5)

## Key Changes

- **`nac_test/utils/controller.py`** (294 lines changed):
  - Added `CredentialSet` dataclass with `env_vars`, `label`, and `auth_method` attributes
  - Added `credential_sets` (ordered list)
  - Removed `CREDENTIAL_PATTERNS` backward-compatible dictionary
  - Removed `CredentialSetStatus` TypedDict
  - SDWAN registry entry now has two credential sets: API Token (20.18+, `auth_method="token"`) and Username/Password (`auth_method="session"`)
  - `_find_credential_sets()`: simplified to track `has_any_var` boolean instead of per-variable present/missing lists; returns `tuple[list[str], list[str], dict[str, CredentialSet]]`
  - `detect_controller_type()`: stores winning `CredentialSet` in module-level `_matched_credential_sets` cache; incomplete error messages now list all accepted credential sets
  - Added `get_matched_credential_set()` — public API for nac-test-pyats-common to retrieve the winning credential set and determine `auth_method`
- **`nac_test/utils/environment.py`** (33 lines changed):
  - `validate_controller_env()`: checks credential sets with early return before requiring username/password
- **`tests/utils/test_controller.py`** (227 lines changed):
  - 43 unit tests covering credential set matching, multi-set priority, partial detection, IOSXE alternative URLs, error formatting, and `get_matched_credential_set()` API
  - Updated assertions from `"missing X_VAR"` → `"incomplete credentials"` format
- **`tests/unit/pyats_core/common/test_base_test_iosxe_credentials.py`**:
  - Updated 1 assertion to match new "incomplete credentials" error format
- **`dev-docs/PRD_AND_ARCHITECTURE.md`** (137 lines changed):
  - Removed stale `CREDENTIAL_PATTERNS` definition
  - Updated detection algorithm pseudocode (`partial_sets` → `partial_controllers`)
  - Updated flow diagram node from "Track best partial match" → "Track as partial if any env_var was set"
- **`CHANGELOG.md`**: Added credential set support entry
- **`README.md`**: Updated credential documentation

## Testing Done

- [ ] Integration tests performed

### Test Commands Used

```bash
# 971 unit tests pass (917 unit + 54 utils/test_controller.py)
cd nac-test && python -m pytest tests/utils/test_controller.py tests/unit/ -x -q

# Verified token credential set detection
export SDWAN_URL=https://x.x.x.x
export SDWAN_API_TOKEN=<jwt_token>
nac-test -d standard -d standard_2015 -d defaults.yaml -f jinja_filters -t test_aaa_2018_token_auth -o results-robot
nac-test -d standard -d standard_2015 -d defaults.yaml -f jinja_filters -t tests -o result-pyats --pyats

# Verified credential set matching
python -c "
from nac_test.utils.controller import detect_controller_type, get_matched_credential_set
ct = detect_controller_type()
cs = get_matched_credential_set(ct)
print(f'{ct}: {cs.label} (auth_method={cs.auth_method})')
"
# Output: SDWAN: API Token (20.18+) (auth_method=token)
```

## Checklist

- [x] Documentation updated (if applicable)
- [ ] No new warnings introduced
- [x] CHANGELOG.md updated (if applicable)

## Additional Notes

- The _matched_credential_sets module-level dict is re-populated in each PyATS subprocess via NACTestBase.setup() → detect_controller_type(). The env vars (which are inherited across subprocess boundaries) are the real source of truth; the dict is an in-process cache.
- Credential set order in CONTROLLER_REGISTRY determines priority: for SDWAN, token auth is checked first (newer method for 20.18+), falling back to username/password session auth. Users can force session auth by unsetting SDWAN_API_TOKEN.

## Additional Testing Done

**PYATS Tests :-** 

1. No env vars
```
ERROR - No controller credentials found in environment

❌ Controller detection failed:
No controller credentials found in environment.

Controller credentials are required for ALL test types (API and D2D).
The framework uses these to determine the architecture context.

Please set environment variables for ONE of the following controller types:

ACI:
  export ACI_URL=<value>
  export ACI_USERNAME=<value>
  export ACI_PASSWORD=<value>

SDWAN:
  (API Token (20.18+)):
  export SDWAN_URL=<value>
  export SDWAN_API_TOKEN=<value>
  Or
  (Username/Password):
  export SDWAN_URL=<value>
  export SDWAN_USERNAME=<value>
  export SDWAN_PASSWORD=<value>

CC:
  export CC_URL=<value>
  export CC_USERNAME=<value>
  export CC_PASSWORD=<value>

MERAKI:
  export MERAKI_URL=<value>
  export MERAKI_USERNAME=<value>
  export MERAKI_PASSWORD=<value>

FMC:
  export FMC_URL=<value>
  export FMC_USERNAME=<value>
  export FMC_PASSWORD=<value>

ISE:
  export ISE_URL=<value>
  export ISE_USERNAME=<value>
  export ISE_PASSWORD=<value>

IOSXE:
  export IOSXE_URL=<value>

Example for ACI:
  export ACI_URL=https://apic.example.com
  export ACI_USERNAME=admin
  export ACI_PASSWORD=yourpassword

Note: Set credentials for only ONE controller type at a time.
```
2. Only SDWAN URL:
```
❌ Controller detection failed:
Incomplete controller credentials detected:
  - SDWAN: incomplete credentials
    Accepted credential sets:
      - API Token (20.18+): SDWAN_URL + SDWAN_API_TOKEN
      - Username/Password: SDWAN_URL + SDWAN_USERNAME + SDWAN_PASSWORD

Please provide ALL required environment variables for your controller type.
```
3. Only SDWAN URL and User
```
ERROR - Incomplete credentials: ['SDWAN']

❌ Controller detection failed:
Incomplete controller credentials detected:
  - SDWAN: incomplete credentials
    Accepted credential sets:
      - API Token (20.18+): SDWAN_URL + SDWAN_API_TOKEN
      - Username/Password: SDWAN_URL + SDWAN_USERNAME + SDWAN_PASSWORD

Please provide ALL required environment variables for your controller type.
```
4. Only SDWAN URL and pass
```
ERROR - Incomplete credentials: ['SDWAN']

❌ Controller detection failed:
Incomplete controller credentials detected:
  - SDWAN: incomplete credentials
    Accepted credential sets:
      - API Token (20.18+): SDWAN_URL + SDWAN_API_TOKEN
      - Username/Password: SDWAN_URL + SDWAN_USERNAME + SDWAN_PASSWORD

Please provide ALL required environment variables for your controller type.
```
5. SDWAN URL, user and wrong pass
Failed testcases

6. Only SDWAN URL, user and right pass
Passed testcases

7. Only SDWAN URL and right token
Pass testcases

8. Only SDWAN URL and wrong token header
Failed testcases

9. Only SDWAN URL and wrong token payload
Failed testcases

10. Only SDWAN URL and wrong token signature
Failed testcases

11. User, SDWAN URL, right password and right token
Passed testcases

12. User, SDWAN URL, wrong pass and right token
Passed testcases

13. User, SDWAN URL, right password and wrong token
Failed testcases

**Robot tests :-**

14. No env var
Failure

15. Only SDWAN URL
Failure

16. Only user and SDWAN URL
Failure

17. Only user, good  pass and SDWAN URL 
Pass

18. Only user, bad pass and SDWAN URL
Failure

19. Only pass and SDWAN URL
Failure

20. SDWAN URL and good token
Pass

21. SDWAN URL and wrong token header
Failure

22. SDWAN URL and wrong token payload
Failure

23. SDWAN URL and wrong token signature
Failure

24. SDWAN URL and user, wrong pass and right token
pass

25. SDWAN URL and user, right pass and wrong token
Fail

26. SDWAN URL and user, right pass and right token
Pass

### DIFF
--- a/nac_test/utils/controller.py
+++ b/nac_test/utils/controller.py
@@ -226,23 +226,7 @@ def detect_controller_type() -> ControllerTypeKey:
 
     # Check for incomplete credentials
     if not complete_sets and partial_controllers:
-        lines_parts: list[str] = []
-        for controller in partial_controllers:
-            config = CONTROLLER_REGISTRY[controller]
-            set_descriptions = [
-                f"{cs.label}: {' + '.join(cs.env_vars)}"
-                for cs in config.credential_sets
-            ]
-            line = f"{controller}: incomplete credentials"
-            line += "\n    Accepted credential sets:\n"
-            line += "\n".join(f"      - {desc}" for desc in set_descriptions)
-            lines_parts.append(line)
-        lines = "\n".join(f"  - {info}" for info in lines_parts)
-        error_message = (
-            f"Incomplete controller credentials detected:\n"
-            f"{lines}\n\n"
-            f"Please provide ALL required environment variables for your controller type."
-        )
+        error_message = _format_incomplete_credentials_error(partial_controllers)
         logger.error(f"Incomplete credentials: {partial_controllers}")
         raise ValueError(error_message)
 
@@ -306,6 +290,45 @@ def _find_credential_sets() -> tuple[list[str], list[str], dict[str, CredentialS
             partial_controllers.append(controller_type)
 
     return complete_sets, partial_controllers, matched_creds
+
+
+def _format_incomplete_credentials_error(partial_controllers: list[str]) -> str:
+    """Format error message for incomplete controller credentials.
+
+    Creates a detailed error message listing each partially configured
+    controller and its accepted credential sets, so the user knows
+    exactly which variables are needed.
+
+    Args:
+        partial_controllers: List of controller types with partial credentials.
+
+    Returns:
+        Formatted error message with accepted credential sets.
+
+    Example:
+        >>> error = _format_incomplete_credentials_error(["SDWAN"])
+        >>> print(error)
+        Incomplete controller credentials detected:
+        ...
+    """
+    lines_parts: list[str] = []
+    for controller in partial_controllers:
+        config = CONTROLLER_REGISTRY[controller]
+        set_descriptions = [
+            f"{cs.label}: {' + '.join(cs.env_vars)}"
+            for cs in config.credential_sets
+        ]
+        line = f"{controller}: incomplete credentials"
+        line += "\n    Accepted credential sets:\n"
+        line += "\n".join(f"      - {desc}" for desc in set_descriptions)
+        lines_parts.append(line)
+    lines = "\n".join(f"  - {info}" for info in lines_parts)
+    return (
+        f"Incomplete controller credentials detected:\n"
+        f"{lines}\n\n"
+        f"Please provide all required variables for one of the "
+        f"accepted credential sets listed above."
+    )
 
 
 def _format_multiple_credentials_error(controllers: list[str]) -> str:
--- a/nac_test/utils/controller.py
+++ b/nac_test/utils/controller.py
@@ -18,6 +18,7 @@ the merged NAC data model.
 
 import logging
 import os
+from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import cast
 
@@ -215,7 +216,9 @@ def detect_controller_type() -> ControllerTypeKey:
     # Check for multiple complete credential sets
     if len(complete) > 1:
         error_message = _format_multiple_credentials_error(list(complete.keys()))
-        logger.error(f"Multiple controller credentials detected: {list(complete.keys())}")
+        logger.error(
+            f"Multiple controller credentials detected: {list(complete.keys())}"
+        )
         raise ValueError(error_message)
 
     # Check for no credentials at all
@@ -292,7 +295,7 @@ def _find_credential_sets() -> tuple[
     return complete, partial
 
 
-def _format_incomplete_credentials_error(partial_controllers: list[str]) -> str:
+def _format_incomplete_credentials_error(partial_controllers: Sequence[str]) -> str:
     """Format error message for incomplete controller credentials.
 
     Creates a detailed error message listing each partially configured
@@ -315,8 +318,7 @@ def _format_incomplete_credentials_error(partial_controllers: list[str]) -> str:
     for controller in partial_controllers:
         config = CONTROLLER_REGISTRY[controller]
         set_descriptions = [
-            f"{cs.label}: {' + '.join(cs.env_vars)}"
-            for cs in config.credential_sets
+            f"{cs.label}: {' + '.join(cs.env_vars)}" for cs in config.credential_sets
         ]
         line = f"{controller}: incomplete credentials"
         line += "\n    Accepted credential sets:\n"
--- a/nac_test/utils/controller.py
+++ b/nac_test/utils/controller.py
@@ -61,8 +61,6 @@ class ControllerConfig:
             (e.g., "defaults.apic", "defaults.sdwan").
         cache_key: The controller_type string passed to AuthCache by the auth adapter.
             None for controllers that don't have an auth adapter in nac-test-pyats-common.
-        alt_url_env_vars: Alternative environment variable names for the URL.
-            Used when a controller supports multiple URL env var names (e.g., IOSXE_HOST).
     """
 
     display_name: str
@@ -71,7 +69,6 @@ class ControllerConfig:
     credential_sets: list[CredentialSet]
     defaults_prefix: str
     cache_key: str | None = None
-    alt_url_env_vars: list[str] | None = None
 
 
 # Single source of truth for all controller configurations
@@ -167,9 +164,12 @@ CONTROLLER_REGISTRY: dict[str, ControllerConfig] = {
                 env_vars=["IOSXE_URL"],
                 label="Device URL",
             ),
+            CredentialSet(
+                env_vars=["IOSXE_HOST"],
+                label="Device Host",
+            ),
         ],
         defaults_prefix="defaults.iosxe",
-        alt_url_env_vars=["IOSXE_HOST"],  # Alternative env var for URL
     ),
 }
 
@@ -270,9 +270,6 @@ def _find_credential_sets() -> tuple[list[str], list[str], dict[str, CredentialS
     complete. If no set is fully satisfied but at least one variable from any set
     is present, the controller is reported as partial.
 
-    This function also handles alternative URL environment variables (e.g., IOSXE_HOST
-    as an alternative to IOSXE_URL) when configured in the ControllerConfig.
-
     Returns:
         A tuple containing:
             - List of controller types with complete credentials
@@ -296,21 +293,7 @@ def _find_credential_sets() -> tuple[list[str], list[str], dict[str, CredentialS
                     has_any_var = True
                     logger.debug(f"  {controller_type}: Found {var}")
                 else:
-                    # Check alternative URL env vars if this is the URL variable
-                    alt_found = False
-                    if var == config.url_env_var and config.alt_url_env_vars:
-                        for alt_var in config.alt_url_env_vars:
-                            alt_value = os.environ.get(alt_var)
-                            if alt_value and alt_value.strip():
-                                has_any_var = True
-                                logger.debug(
-                                    f"  {controller_type}: Found {alt_var} (alternative)"
-                                )
-                                alt_found = True
-                                break
-
-                    if not alt_found:
-                        all_present = False
+                    all_present = False
 
             if all_present:
                 complete_sets.append(controller_type)
@@ -491,9 +474,8 @@ def get_defaults_prefix(controller_type: str) -> str:
 def get_controller_url(controller_type: str) -> str:
     """Get the controller URL from environment variables.
 
-    Looks up the primary URL environment variable from CONTROLLER_REGISTRY, and
-    also checks alternative URL env vars if configured (e.g., IOSXE_HOST as an
-    alternative to IOSXE_URL).
+    Iterates through credential sets in order, returning the first env var value
+    found. This follows the same first-match-wins pattern as _find_credential_sets.
 
     Args:
         controller_type: The internal controller type key (e.g., "ACI", "SDWAN", "IOSXE").
@@ -502,7 +484,7 @@ def get_controller_url(controller_type: str) -> str:
         The controller URL value from the environment.
 
     Raises:
-        KeyError: If neither the primary nor any alternative URL env var is set.
+        KeyError: If no credential set env var has a URL value set.
 
     Example:
         >>> os.environ["ACI_URL"] = "https://apic.example.com"
@@ -519,17 +501,12 @@ def get_controller_url(controller_type: str) -> str:
         # Fallback for unknown controller types
         return os.environ[f"{controller_type}_URL"]
 
-    # Try primary URL env var first
-    url_value = os.environ.get(config.url_env_var)
-    if url_value and url_value.strip():
-        return url_value.strip()
-
-    # Try alternative URL env vars if configured
-    if config.alt_url_env_vars:
-        for alt_var in config.alt_url_env_vars:
-            alt_value = os.environ.get(alt_var)
-            if alt_value and alt_value.strip():
-                return alt_value.strip()
+    # Iterate credential sets — first env var with a value wins
+    for cred_set in config.credential_sets:
+        for var in cred_set.env_vars:
+            value = os.environ.get(var)
+            if value and value.strip():
+                return value.strip()
 
     # No URL found - raise KeyError with the primary var name
     raise KeyError(config.url_env_var)
--- a/nac_test/utils/controller.py
+++ b/nac_test/utils/controller.py
@@ -39,6 +39,23 @@ class CredentialSetStatus(TypedDict):
 
 
 @dataclass(frozen=True)
+class CredentialSet:
+    """A single credential combination that can authenticate to a controller.
+
+    Each set is self-contained: if ALL env_vars are present and non-empty,
+    the controller is considered fully configured. When a controller has
+    multiple CredentialSets, the first satisfied set wins (order matters).
+
+    Attributes:
+        env_vars: Environment variable names required for this credential method.
+        label: Human-readable label for error messages (e.g., "API Token (20.18+)").
+    """
+
+    env_vars: list[str]
+    label: str
+
+
+@dataclass(frozen=True)
 class ControllerConfig:
     """Configuration metadata for a supported controller type.
 
@@ -46,7 +63,9 @@ class ControllerConfig:
         display_name: User-facing name (e.g., "APIC", "Catalyst Center").
         url_env_var: Environment variable name for the controller URL.
         env_var_prefix: Prefix for credential env vars (e.g., "ACI" → ACI_USERNAME).
-        required_env_vars: List of environment variables required for this controller.
+        credential_sets: Ordered list of credential combinations. The first set
+            whose env_vars are all present and non-empty wins. Every controller
+            must have at least one CredentialSet.
         defaults_prefix: JMESPath prefix for the defaults block in NAC data models
             (e.g., "defaults.apic", "defaults.sdwan").
         cache_key: The controller_type string passed to AuthCache by the auth adapter.
@@ -58,7 +77,7 @@ class ControllerConfig:
     display_name: str
     url_env_var: str
     env_var_prefix: str
-    required_env_vars: list[str]
+    credential_sets: list[CredentialSet]
     defaults_prefix: str
     cache_key: str | None = None
     alt_url_env_vars: list[str] | None = None
@@ -71,7 +90,12 @@ CONTROLLER_REGISTRY: dict[str, ControllerConfig] = {
         display_name="APIC",
         url_env_var="ACI_URL",
         env_var_prefix="ACI",
-        required_env_vars=["ACI_URL", "ACI_USERNAME", "ACI_PASSWORD"],
+        credential_sets=[
+            CredentialSet(
+                env_vars=["ACI_URL", "ACI_USERNAME", "ACI_PASSWORD"],
+                label="Username/Password",
+            ),
+        ],
         defaults_prefix="defaults.apic",
         cache_key="ACI",
     ),
@@ -79,7 +103,16 @@ CONTROLLER_REGISTRY: dict[str, ControllerConfig] = {
         display_name="SDWAN Manager",
         url_env_var="SDWAN_URL",
         env_var_prefix="SDWAN",
-        required_env_vars=["SDWAN_URL", "SDWAN_USERNAME", "SDWAN_PASSWORD"],
+        credential_sets=[
+            CredentialSet(
+                env_vars=["SDWAN_URL", "SDWAN_API_TOKEN"],
+                label="API Token (20.18+)",
+            ),
+            CredentialSet(
+                env_vars=["SDWAN_URL", "SDWAN_USERNAME", "SDWAN_PASSWORD"],
+                label="Username/Password",
+            ),
+        ],
         defaults_prefix="defaults.sdwan",
         cache_key="SDWAN_MANAGER",
     ),
@@ -87,7 +120,12 @@ CONTROLLER_REGISTRY: dict[str, ControllerConfig] = {
         display_name="Catalyst Center",
         url_env_var="CC_URL",
         env_var_prefix="CC",
-        required_env_vars=["CC_URL", "CC_USERNAME", "CC_PASSWORD"],
+        credential_sets=[
+            CredentialSet(
+                env_vars=["CC_URL", "CC_USERNAME", "CC_PASSWORD"],
+                label="Username/Password",
+            ),
+        ],
         defaults_prefix="defaults.catc",
         cache_key="CC",
     ),
@@ -95,38 +133,58 @@ CONTROLLER_REGISTRY: dict[str, ControllerConfig] = {
         display_name="Meraki",
         url_env_var="MERAKI_URL",
         env_var_prefix="MERAKI",
-        required_env_vars=["MERAKI_URL", "MERAKI_USERNAME", "MERAKI_PASSWORD"],
+        credential_sets=[
+            CredentialSet(
+                env_vars=["MERAKI_URL", "MERAKI_USERNAME", "MERAKI_PASSWORD"],
+                label="Username/Password",
+            ),
+        ],
         defaults_prefix="defaults.meraki",
     ),
     "FMC": ControllerConfig(
         display_name="Firepower Management Center",
         url_env_var="FMC_URL",
         env_var_prefix="FMC",
-        required_env_vars=["FMC_URL", "FMC_USERNAME", "FMC_PASSWORD"],
+        credential_sets=[
+            CredentialSet(
+                env_vars=["FMC_URL", "FMC_USERNAME", "FMC_PASSWORD"],
+                label="Username/Password",
+            ),
+        ],
         defaults_prefix="defaults.fmc",
     ),
     "ISE": ControllerConfig(
         display_name="ISE",
         url_env_var="ISE_URL",
         env_var_prefix="ISE",
-        required_env_vars=["ISE_URL", "ISE_USERNAME", "ISE_PASSWORD"],
+        credential_sets=[
+            CredentialSet(
+                env_vars=["ISE_URL", "ISE_USERNAME", "ISE_PASSWORD"],
+                label="Username/Password",
+            ),
+        ],
         defaults_prefix="defaults.ise",
     ),
     "IOSXE": ControllerConfig(
         display_name="IOS XE",
         url_env_var="IOSXE_URL",
         env_var_prefix="IOSXE",
-        required_env_vars=[
-            "IOSXE_URL"
-        ],  # Direct device access, no controller credentials
+        # Direct device access, no controller credentials required
+        credential_sets=[
+            CredentialSet(
+                env_vars=["IOSXE_URL"],
+                label="Device URL",
+            ),
+        ],
         defaults_prefix="defaults.iosxe",
         alt_url_env_vars=["IOSXE_HOST"],  # Alternative env var for URL
     ),
 }
 
 # Backward compatibility - remove in future version
+# Uses the first credential set's env_vars as the "primary" pattern
 CREDENTIAL_PATTERNS: dict[str, list[str]] = {
-    controller_type: config.required_env_vars
+    controller_type: config.credential_sets[0].env_vars
     for controller_type, config in CONTROLLER_REGISTRY.items()
 }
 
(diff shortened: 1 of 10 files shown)


Anush-Mangaluru:
### Documenting code review history for reference (token support + refactoring):  

**1. Thought process shared by Oliver :** 

- We define a credential set in nac-test
- We iterate through the elements and take the first one which wins. We remember which one was used
- We define an API in nac-test for nac-test-pyats-common to retrieve the credential element which was found
- nac-test-pyats-common uses it and does the needful (possibly we define some attribute in the credential element which nac-test-pyats-common can use to tell which auth mechanism to use.
	
**2. First set of review comments and address :** 

- Can we drop CREDENTIAL_PATTERNS as part of this refactor.
		_[AK] - Yes this is done now_
- Maintaining _matched_credential_sets at module level looks fragile, bearing in mind that the pyats processes are running in subprocesses and wouldn't be able to access this? Maybe we will need to do another credential detection in nac-test-pyats-common code?
		_[AK] - This is what I understood analysing codebase with AI, i.e. NACTestBase.setup() (decorated with @aetest.setup) calls detect_controller_type() automatically before any test method runs. This reads the inherited environment variables and re-populates the dict in that subprocess before SDWANManagerAuth.get_auth() calls get_matched_credential_set() . So, we need not perform a credential detection again._
                 _[OB] - Ok, so we do check credentials then  twice: once in pre-flight check, and then once per pyats test subprocess._
- I understand you are doing some heuristics, i.e. check for all env vars which look like controller env vars, and then perform a "best match".  Even if this is less efficient, I would rather have us look at the sets in order, possibly at the expense that we're no longer able to tell which vars are missing exactly, which is not that important-a-feature
		_[AK] - This makes sense, removed unnecessary heuristics_
- And it would also be good to be deterministic (and also describe the preference somewhere).. so what happens when a user as USER/PASSWORD/TOKEN in the repo, the selection must be deterministic, so you would need to define a preference
		_[AK] - My thought here is that selection is already deterministic — credential sets are defined in a fixed order in CONTROLLER_REGISTRY, and the first fully-satisfied set wins. For SDWAN, token auth is checked first since it's the newer method (20.18+), falling back to username/password session auth. If a user has both defined and wants to force session auth, they simply unset SDWAN_API_TOKEN. Adding an explicit preference mechanism would add unnecessary complexity without clear benefit_

**3.  Oliver's agent review findings and address :** 

- [HIGH] nac_test/utils/environment.py — validate_controller_env() checks env var presence with all(os.environ.get(v, "").strip() ...) and does not honor alt_url_env_vars (e.g. IOSXE IOSXE_HOST), while _find_credential_sets() does honor alt URL vars. Why this matters: you can pass detection via IOSXE_HOST, but validate_controller_env() may still fail later because it only checks IOSXE_URL (or the primary env var in a set). Suggested fix: reuse the same URL-var check logic as _find_credential_sets() (treat config.url_env_var as satisfied if any alt_url_env_vars is set), or factor out a shared helper in [controller.py](https://controller.py/) used by both.
	_[AK] Good catch. Removed alt_url_env_vars and used credset for this usecase. Tested with unittests to ensure IOSXE vars work_
	
- [MED] tests/utils/test_controller.py — Several “incomplete credentials” tests no longer assert enough about which controller(s) were considered partial. Example: test_url_only_is_partial only asserts the generic header, but not that the message includes SDWAN: incomplete credentials or the accepted credential sets; a regression could make the error about a different controller and still pass. Suggested fix: in those tests, assert the controller line and accepted set lines are present (they already do this in test_incomplete_error_shows_credential_set_options, so mirror that strength elsewhere). Error handling.
	_[AK] Strengthened test_url_only_is_partial assertions to match other partial-credential tests_

- [MED] nac_test/utils/environment.py — It imports CONTROLLER_REGISTRY inside the function (from nac_test.utils.controller import CONTROLLER_REGISTRY). Not necessarily wrong, but it increases coupling and can create import-cycle fragility if controller/environment evolve. Suggested fix: if an import cycle is the concern, consider importing controller module and referencing controller.CONTROLLER_REGISTRY (same net effect but clearer), or move registry access behind a minimal accessor.
	_[AK] Agreed it's worth watching. The function-level import is intentional to avoid a potential circular dependency (environment.py ↔ controller.py). Currently it's one-way, so no cycle exists today. Would be good to revisit later if the modules evolve and coupling grows._

- [MED] nac_test/utils/environment.py — Credential-set satisfaction is now implemented twice: once in controller detection (_find_credential_sets()), and again in validate_controller_env(), but with different semantics (notably alt url handling, and “first set is default” reporting). Suggested fix: make validate_controller_env() call into a shared function (even a small one) to decide “is any set satisfied?”, and optionally to compute “missing vars” for the message.
	_[AK] This was a valid concern when alt_url_env_vars existed — the two paths had different semantics. With the removal of alt_url_env_vars, both now iterate credential_sets identically. The core check is a one-liner (all(os.environ.get(v, "").strip() ...), so extracting a shared helper would add indirection without meaningful benefit._

- [MED] nac_test/utils/environment.py — Now reaches into controller registry (and re-implements set checking). That’s a layering smell: [environment.py](https://environment.py/) is becoming controller-registry-aware, while [controller.py](https://controller.py/) already owns “what constitutes a valid credential configuration.” Suggested fix: make environment validation call a controller utility (e.g., “is any cred set satisfied?” / “missing vars for message”), rather than duplicating registry logic.
	_[AK] Same reasoning as above — with alt_url_env_vars removed, the check is a trivial one-liner over credential_sets. Creating dedicated APIs in controller.py for a single consumer would add indirection without reducing complexity._

- [LOW] nac_test/utils/controller.py — _format_multiple_credentials_error() defines an inner helper _all_env_vars() each call. This is fine, but it’s also essentially “registry → union of env vars per controller” logic that may be useful elsewhere. Suggested fix: only if it repeats elsewhere; otherwise OK to keep local.
	_[AK] It doesn't repeat anywhere. It's a local helper used once, inside one error-formatting function. Keep it as-is, no action needed._

- [LOW] tests/utils/test_controller.py — Many classes repeat the same “clean environment” fixture logic iterating through CONTROLLER_REGISTRY credential sets and popping env vars. This is now copy-pasted in multiple places in the file. Suggested fix: a single helper/fixture in this test module (or a local function) to clear controller vars + optionally clear _matched_credential_sets.
	_[AK] Consolidated 6 per-class clean_environment fixtures into single module-level fixture_

- [LOW] nac_test/utils/controller.py — get_matched_credential_set(controller_type: str) takes a str even though only specific controller keys are valid. This leaks a weaker contract and makes call sites less self-documenting. Suggested fix: accept ControllerTypeKey (or the same type returned by detect_controller_type()).
	_[AK] Agree — detect_controller_type() already returns ControllerTypeKey. Keeping str here avoids forcing a cross-package type import on nac-test-pyats-common (the main caller). The function returns None gracefully for unknown keys, so no runtime risk._

- [LOW] nac_test/utils/controller.py — _matched_credential_sets: dict[str, CredentialSet] uses str keys while detect_controller_type() returns ControllerTypeKey and the registry keys are effectively ControllerTypeKey. It’s not wrong, but it weakens type precision. Suggested fix: consider dict[ControllerTypeKey, CredentialSet] if ControllerTypeKey is a Literal[...]/alias; also adjust get_matched_credential_set() signature to accept ControllerTypeKey if that’s the intended API.
	_[AK] Same as above — the dict is private, writes go through detect_controller_type() which already uses ControllerTypeKey, so type safety is enforced at the write side_

nac_test/utils/controller.py
```diff
@@ -19,23 +19,31 @@
...
-        missing: List of environment variable names that are not set.
+        env_vars: Environment variable names required for this credential method.
+        label: Human-readable label for error messages (e.g., "API Token (20.18+)").
+        auth_method: Identifier consumed by auth adapters in nac-test-pyats-common
+            to select the authentication mechanism (e.g., "token", "session").
     """
 
-    present: list[str]
-    missing: list[str]
+    env_vars: list[str]
```
oboehmer: **[Medium]** `frozen=True` dataclass with mutable `list` field: The dataclass is frozen (can't reassign fields), but the list inside is still mutable — someone could do `cred_set.env_vars.append("NEW_VAR")` and silently break invariants.

Use `tuple[str, ...]` for `env_vars` and `tuple[CredentialSet, ...]` for `credential_sets` in `ControllerConfig` to match the frozen intent. Single-line change per field.

nac_test/utils/controller.py
```diff
@@ -172,102 +219,93 @@ def detect_controller_type() -> ControllerTypeKey:
...
+            ]
+            line = f"{controller}: incomplete credentials"
+            line += "\n    Accepted credential sets:\n"
+            line += "\n".join(f"      - {desc}" for desc in set_descriptions)
+            lines_parts.append(line)
+        lines = "\n".join(f"  - {info}" for info in lines_parts)
         error_message = (
             f"Incomplete controller credentials detected:\n"
             f"{lines}\n\n"
             f"Please provide ALL required environment variables for your controller type."
```
oboehmer: This message is now misleading — we need all vars from ONE credential set, not all vars across all sets.

Suggested wording:
> "Please provide all required variables for one of the accepted credential sets listed above."

Also: this inline error formatting should be extracted to a `_format_incomplete_credentials_error(partial_controllers)` helper for consistency with the existing `_format_no_credentials_error` and `_format_multiple_credentials_error` pattern.

nac_test/utils/controller.py
```diff
@@ -172,102 +219,93 @@ def detect_controller_type() -> ControllerTypeKey:
...
+        logger.error(f"Incomplete credentials: {partial_controllers}")
         raise ValueError(error_message)
 
     # Exactly one complete set found - success
     # complete_sets come from CONTROLLER_REGISTRY keys, which are always valid
     # ControllerTypeKey values, but mypy can't infer this from dict iteration.
     controller_type = cast(ControllerTypeKey, complete_sets[0])
-    logger.info(f"Detected controller type: {controller_type}")
+    # Store the winning credential set for retrieval by auth adapters
+    if controller_type in matched_creds:
```
oboehmer: **Dead code:** The `else` branch (line 261) is unreachable. If a controller is in `complete_sets`, it is *always* in `matched_creds` — both are populated in the same `if all_present` block inside `_find_credential_sets()`. This motivates the refactor below.

---

**Suggested refactor:** `_find_credential_sets()` returns a 3-tuple that forces the caller to "join" a list with a dict, and the defensive `if controller_type in matched_creds` check (which is always True) makes the logic harder to grasp.

Clearer if `_find_credential_sets()` returned **complete controllers as a dict**:

```python
def _find_credential_sets() -> tuple[
    dict[ControllerTypeKey, CredentialSet],
    list[ControllerTypeKey],
]:
    ...
```

Where:
- `complete` maps `controller_type -> winning CredentialSet` (first satisfied set wins)
- `partial` is a list of controllers with `has_any_var` but no set fully satisfied

Then `detect_controller_type()` becomes:

```python
complete, partial = _find_credential_sets()

if len(complete) > 1:
    raise ...
if not complete and not partial:
    raise ...
if not complete and partial:
    raise ...

controller_type = next(iter(complete))
_matched_credential_sets[controller_type] = complete[controller_type]
return controller_type
```

Mechanical refactor (no behavior change) — makes the invariant "complete implies winning set exists" explicit in the type.

nac_test/utils/controller.py
```diff
@@ -172,102 +219,93 @@ def detect_controller_type() -> ControllerTypeKey:
...
-                "missing": missing_vars,
-            }
-
-    return complete_sets, partial_sets
+        for cred_set in config.credential_sets:
+            all_present = True
+
+            for var in cred_set.env_vars:
+                value = os.environ.get(var)
+                if value and value.strip():
```
oboehmer: **[Medium / DRY]** The "is env var set and non-empty?" check pattern appears in three places with slightly different coding styles:

| Location | Pattern |
|----------|--------|
| `_find_credential_sets` (here) | `value = os.environ.get(var); if value and value.strip()` |
| `validate_controller_env` | `all(os.environ.get(v, "").strip() for v in ...)` |
| `get_controller_url` | `value = os.environ.get(var); if value and value.strip()` |

They're semantically equivalent but the different styles invite drift. Extract:
```python
def _is_env_var_set(var: str) -> bool:
    """Check if env var exists and has a non-whitespace value."""
    value = os.environ.get(var)
    return bool(value and value.strip())
```

This also makes the intent self-documenting. Can be deferred to #856.

nac_test/utils/controller.py
```diff
@@ -452,17 +501,39 @@ def get_controller_url(controller_type: str) -> str:
...
-        return url_value.strip()
-
-    # Try alternative URL env vars if configured
-    if config.alt_url_env_vars:
-        for alt_var in config.alt_url_env_vars:
-            alt_value = os.environ.get(alt_var)
-            if alt_value and alt_value.strip():
-                return alt_value.strip()
+    # Iterate credential sets — first env var with a value wins
+    for cred_set in config.credential_sets:
```
oboehmer, line 535: **[High]** Semantic issue: this function iterates ALL `env_vars` from ALL credential sets and returns the first non-empty value. It relies on an **unwritten convention** that the URL is always the first var listed in `env_vars`.

For SDWAN, credential set 1 is `["SDWAN_URL", "SDWAN_API_TOKEN"]`. If `SDWAN_URL` were empty but `SDWAN_API_TOKEN` set, this would return the token as the "URL". While this can't happen in normal flow (detection ensures all vars are set), `get_controller_url` is a public API.

Also note: there's a separate `_get_controller_url` in `cli/validators/controller_auth.py` (line 61-74) that uses `config.url_env_var` directly — two implementations with different semantics.

Suggested fix for this PR:
```python
def get_controller_url(controller_type: str) -> str:
    config = CONTROLLER_REGISTRY.get(controller_type)
    if config is None:
        return os.environ[f"{controller_type}_URL"]

    # Primary URL
    value = os.environ.get(config.url_env_var, "").strip()
    if value:
        return value

    # Check alternative credential sets (IOSXE_HOST case)
    for cred_set in config.credential_sets:
        first_var = cred_set.env_vars[0]
        if first_var != config.url_env_var:
            alt = os.environ.get(first_var, "").strip()
            if alt:
                return alt

    raise KeyError(config.url_env_var)
```

Longer term (#856): unify with `_get_controller_url` in controller_auth.py, and consider caching the resolved URL in a `DetectionResult` dataclass to avoid repeated env scanning.

nac_test/utils/environment.py
```diff
@@ -122,19 +122,36 @@ def get_int(var_name: str, default: int = 0) -> int:
...
+        config = CONTROLLER_REGISTRY.get(controller_type)
+        if config:
+            # Check each credential set — first fully satisfied wins
+            for cred_set in config.credential_sets:
+                if all(os.environ.get(v, "").strip() for v in cred_set.env_vars):
+                    return  # Credentials satisfied
+
+            # No set fully satisfied — report missing vars from first set
+            # (the "default" credential method)
+            required_vars = config.credential_sets[0].env_vars
```
oboehmer: **[High]** Inconsistent error reporting: when no credential set is fully satisfied, this reports missing vars from **only the first credential set** (token set for SDWAN). A user intending to use username/password would see "missing SDWAN_API_TOKEN" — confusing.

This is inconsistent with `detect_controller_type()`'s error which lists ALL accepted credential sets.

Options:
1. Match `detect_controller_type()`'s behavior: list all accepted sets
2. Report the set closest to being satisfied (most vars present)
3. At minimum, report the first set that has _any_ partial match

Also: the validation logic here (`all(os.environ.get(v, "").strip() for v in cred_set.env_vars)`) duplicates `_find_credential_sets()` — see DRY comment on controller.py:292.

tests/utils/test_controller.py
```diff
@@ -535,3 +481,133 @@ def test_get_controller_url_unknown_controller_raises_when_not_set(self) -> None
...
+        os.environ["ACI_URL"] = "https://apic.example.com"
+        os.environ["ACI_USERNAME"] = "admin"
+        os.environ["ACI_PASSWORD"] = "password"
+
+        detect_controller_type()
+
+        cred = get_matched_credential_set("ACI")
+        assert cred is not None
+        assert cred.auth_method == "session"
+        assert cred.label == "Username/Password"
```
oboehmer, line 641: **[Low] Test coverage gaps to consider:**

1. **No test for `validate_controller_env` with multi-credential-set controllers** — e.g., does `validate_controller_env("SDWAN")` pass when only username/password is set (bypassing the token set)? Given the inconsistency flagged in environment.py, this matters.

2. **No test for `get_controller_url("SDWAN")`** — the parametrized test sets `SDWAN_URL` and passes, but doesn't exercise the iteration logic when URL is missing. A test with `SDWAN_URL=""` + `SDWAN_API_TOKEN` set would expose the semantic bug flagged on controller.py:505.

tests/utils/test_controller.py
```diff
@@ -535,3 +481,133 @@ def test_get_controller_url_unknown_controller_raises_when_not_set(self) -> None
...
+
+    SDWAN supports two credential methods:
+    1. API Token (20.18+): SDWAN_URL + SDWAN_API_TOKEN  (first — wins when both present)
+    2. Username/Password: SDWAN_URL + SDWAN_USERNAME + SDWAN_PASSWORD
+    """
+
+    def test_detect_sdwan_with_api_token(self) -> None:
+        """Test SDWAN detection with API token credentials."""
+        os.environ["SDWAN_URL"] = "https://vmanage.example.com"
+        os.environ["SDWAN_API_TOKEN"] = "eyJhbGciOiJSUzI1NiJ9.test.sig"
```
oboehmer, line 525: [High]: this pollutes the environment. Please use the fixture mentioned earlier to clean the environment

tests/utils/test_controller.py
```diff
@@ -161,31 +168,19 @@ def test_d2d_scenario_with_dummy_credentials(self) -> None:
 class TestHelperFunctions:
     """Test helper functions for credential detection."""
 
-    @pytest.fixture(autouse=True)
-    def clean_environment(self) -> Generator[None, None, None]:
```
oboehmer, line 165: We still need this, please adjust to new logic/datastructures and use it throughout (maybe through a module-level autouse fixture.

oboehmer (changes requested):
Thanks again for your contribution and the solid refactor to support the new SDWAN auth method. This will provide much more flexibility as we add more architectures/controllers in the future. I have raised another refactoring issue #856 to clean up some tech dept in our code, but there are still a few items  I would like to ask you to address. 

Next to some codebase changes, I would like to re-introduce proper environment cleanup autouse fixture which you removed. The unit tests need to succeed with any controller env vars being populated in the caller, so we need to do proper cleanup before the tests start.

We also need to ensure for proper unit test coverage for all controllers and a better permutation.  I noticed the following bug when setting IOSXE_URL and SDWAN_URL, but I miss IOSXE_USERNAME. The controller detection passes and it selects IOSXE.. so something is still at odds with the logic.
Update: this is due to an incorrect configuration, we should also set IOSXE_USERNAME/PASSWORD.. let me comment there.. 

```
(nac-test) OBOEHMER-M-5D76:sdwan oboehmer$ set|egrep '^(SDWAN|IOSXE)'
IOSXE_PASSWORD=cisco123
IOSXE_URL=foo
SDWAN_URL=http://localhost
(nac-test) OBOEHMER-M-5D76:sdwan oboehmer$ nac-test -d data/ -t tests/ -o /tmp/r --loglevel INFO


📄 Merging data model files...
INFO - Loading yaml files from data
INFO - Writing merged data model to /tmp/r/merged_data_model_test_variables.yaml
✅ Data model merging completed (< 1s)
INFO - Detected controller type: IOSXE (auth_method=session)
INFO - Controller type detected: IOSXE
[...]
````

So having properly designed unit tests which test various combinations is paramount. 

please let me know if you have any concerns or questions.

thanks so much again.

File: nac_test/utils/controller.py [hunk 10]
```diff
@@ -46,89 +54,128 @@ class ControllerConfig:
...
+        credential_sets=[
+            CredentialSet(
+                env_vars=["IOSXE_URL"],
+                label="Device URL",
+            ),
+            CredentialSet(
+                env_vars=["IOSXE_HOST"],
+                label="Device Host",
+            ),
+        ],
```
oboehmer: we missed IOSXE_USERNAME and IOSXE_PASSWORD here (also in the original code), please add them to both sets

```
                env_vars=["IOSXE_URL", "IOSXE_USERNAME", "IOSXE_PASSWORD"],
```

Anush-Mangaluru:
Sure @oboehmer . I have started working on your comments and will respond back once all comments are addressed.
Thank you.

(Note: I am giving thumbs up emoji for comments I have completed working on. I am still WIP, will get all the comments addressed at earliest)

tests/utils/test_controller.py
Anush-Mangaluru: The module level autouse fixture already in place in line:25

tests/utils/test_controller.py
Anush-Mangaluru: The module level autouse fixture already in place in line:25

nac_test/utils/controller.py
Anush-Mangaluru: Updated

nac_test/utils/controller.py
Anush-Mangaluru: Updated

nac_test/utils/controller.py
Anush-Mangaluru: Updated

nac_test/utils/controller.py
Anush-Mangaluru: Updated in this PR itself

nac_test/utils/controller.py
Anush-Mangaluru: Updated the changes requested for this PR.
Longer term as mentioned can be planned in #856

nac_test/utils/environment.py
Anush-Mangaluru: updated

tests/utils/test_controller.py
Anush-Mangaluru: updated

nac_test/utils/controller.py
Anush-Mangaluru: Added the additional fields for IOSXE

Anush-Mangaluru:
> Sure @oboehmer . I have started working on your comments and will respond back once all comments are addressed. Thank you.
> 
> (Note: I am giving thumbs up emoji for comments I have completed working on. I am still WIP, will get all the comments addressed at earliest)

@oboehmer  - Completed requested changes for this PR. Now, will work on nac-test-pyats-common repo comments.
Thanks.

tests/utils/test_controller.py
oboehmer: 😬

oboehmer (approved):
Thanks so much for addressing the changes.. Approved, but maybe you can add one more test to `TestEdgeCases` which asserts a specific condition not yet evaluated

```python
def test_iosxe_partial_and_sdwan_partial_are_both_reported(self) -> None:
    """Regression test: IOSXE_URL + IOSXE_PASSWORD (no IOSXE_USERNAME) combined
    with SDWAN_URL must NOT detect IOSXE — both controllers should be reported
    as partial, not complete.

    Before the fix that added IOSXE_USERNAME/IOSXE_PASSWORD to the IOSXE
    credential sets, setting only IOSXE_URL was enough to satisfy detection,
    so this combination incorrectly returned 'IOSXE' instead of raising.
    """
    os.environ["IOSXE_URL"] = "https://iosxe.example.com"
    os.environ["IOSXE_PASSWORD"] = "cisco123"
    # IOSXE_USERNAME deliberately omitted — credential set must not be satisfied
    os.environ["SDWAN_URL"] = "https://vmanage.example.com"
    # No SDWAN credentials beyond URL

    with pytest.raises(ValueError) as exc_info:
        detect_controller_type()

    error_msg = str(exc_info.value)
    assert "Incomplete controller credentials detected" in error_msg
    assert "IOSXE: incomplete credentials" in error_msg
    assert "SDWAN: incomplete credentials" in error_msg
```

Anush-Mangaluru:
> Thanks so much for addressing the changes.. Approved, but maybe you can add one more test to `TestEdgeCases` which asserts a specific condition not yet evaluated
> 
> ```python
> def test_iosxe_partial_and_sdwan_partial_are_both_reported(self) -> None:
>     """Regression test: IOSXE_URL + IOSXE_PASSWORD (no IOSXE_USERNAME) combined
>     with SDWAN_URL must NOT detect IOSXE — both controllers should be reported
>     as partial, not complete.
> 
>     Before the fix that added IOSXE_USERNAME/IOSXE_PASSWORD to the IOSXE
>     credential sets, setting only IOSXE_URL was enough to satisfy detection,
>     so this combination incorrectly returned 'IOSXE' instead of raising.
>     """
>     os.environ["IOSXE_URL"] = "https://iosxe.example.com"
>     os.environ["IOSXE_PASSWORD"] = "cisco123"
>     # IOSXE_USERNAME deliberately omitted — credential set must not be satisfied
>     os.environ["SDWAN_URL"] = "https://vmanage.example.com"
>     # No SDWAN credentials beyond URL
> 
>     with pytest.raises(ValueError) as exc_info:
>         detect_controller_type()
> 
>     error_msg = str(exc_info.value)
>     assert "Incomplete controller credentials detected" in error_msg
>     assert "IOSXE: incomplete credentials" in error_msg
>     assert "SDWAN: incomplete credentials" in error_msg
> ```

@oboehmer - Updated, thank you :)

oboehmer (approved):
thanks again for this, will merge it now

Issue #679, opened 2026-03-20
fix: replace assert with explicit guard for controller_type narrowing in orchestrator


aitestino:
## Problem

In `nac_test/core/orchestrator.py`, `assert` is used for type narrowing:

```python
assert controller_type is not None
```

`assert` statements are stripped when Python runs with `-O` (optimized mode), which means this guard would silently disappear in production. Using `assert` for control flow or type narrowing in production code is an anti-pattern — it should only be used in tests.

## Proposed Solution

Replace with an explicit guard:

```python
if controller_type is None:
    raise NacTestError("controller_type must not be None at this point")
```

Or use a helper like:

```python
def ensure_not_none(value: T | None, msg: str) -> T:
    if value is None:
        raise NacTestError(msg)
    return value
```

## Context

This was flagged by multiple review agents (best-practices, skeptic, QA) as a code quality issue. The `assert` is doing real work — it narrows the type for mypy — but it needs to be a proper runtime check.

## Acceptance Criteria

- [ ] No `assert` used for type narrowing in production code
- [ ] Replacement raises a proper exception (`NacTestError` or `ValueError`)
- [ ] mypy still passes with the replacement

PR #863, opened 2026-06-09
feat(broker): support Genie-driven command execution in broker mode

### LINKED ISSUE #663: [bug]  Genie BGP Parser Returns Inconsistent VRF Keys Causing False Test Failures

oboehmer:
## Description

Enable Genie-driven command execution through the connection broker. The broker remains a dumb SSH multiplexer — Genie runs client-side, with supplementary commands transparently routed through the broker via IPC.

**Problem:** Some Genie parsers internally call `device.execute()` for supplementary commands (e.g. `show running-config | section router ospf 1` for VRF resolution). In broker mode, the test client's testbed device has no live SSH session — only the broker server holds connections. Without intervention, supplementary commands fail.

**Solution:** Patch the client-side testbed device so it *appears* fully connected to Genie. Any command Genie fires transparently routes through the broker via IPC:

```
Test client (patched device)          Broker server
────────────────────────────          ─────────────
parse_output(cmd)
  → Genie runs parser locally
  → parser calls device.cli.execute("show run | sec ...")
    → _BrokerCliShim.execute()
      → broker_client.execute_command()  ──IPC──→  SSH → output
    ← output (cached for next caller)
  ← parsed dict with supplementary data
```

This keeps the broker as a dumb execution pipe (no Genie dependency), preserves test writer flexibility (`device.parse()` works transparently in both modes), and requires no test template signature changes beyond adding `await`.

## Closes

- Fixes #663

## Related Issue(s)

- #835 (alternative Approach A by @seojumper — broker-side parsing; superseded)

## Type of Change

- [ ] Documentation update

## Test Framework Affected

- [x] PyATS
- [ ] Robot Framework
- [ ] Both

- [ ] ACI (APIC)
- [x] Catalyst SD-WAN (SDWAN Manager / vManage)
- [ ] Meraki (Cloud-managed)
- [x] NX-OS (Nexus Direct-to-Device)
- [x] IOS-XE (Direct-to-Device)
- [x] IOS-XR (Direct-to-Device)
- [ ] Hyperfabric
- [ ] All architectures

## Platform Tested

- [x] macOS (version tested: macOS 15 / Python 3.12)

## Key Changes

**Core feature:**
- `_patch_device_execute_for_broker()`: patches `device.connected`, `connectionmgr.is_connected`, and `device.cli` with `_BrokerCliShim` that delegates to broker
- `parse_output()` made async — runs Genie in worker thread via `run_in_executor` so supplementary broker calls don't deadlock the event loop
- Unified command execution through `testbed_device.execute` in both broker and direct mode (eliminates mode-specific branching)
- Integration test fixtures exercising Genie-driven execution with cache miss/hit validation across files

**Code quality (from @aitestino PR #835 review):**
- `CommandCache.get_cache_stats()`: made thread-safe (snapshot under lock)
- Replace `# type: ignore` with `cast()` for type narrowing (with motivating comments)
- Add `DEVICE_EXECUTE_TIMEOUT` to `constants.__all__`
- Unit tests for Genie broker side-effects (`connected`, `connectionmgr.is_connected`, `cli` shim)
- Fix TITLE copy-paste in `test_broker_genie_parse_2.py`
- Exclude `tests/integration/fixtures` from pytest collection

**Cleanup (addressing dead code from unified execution refactor):**
- Remove dead `connection` parameter from `_create_execute_command_method()` (was unused after the `testbed_device.execute` unification)
- Enforce testbed device invariant at the top of `_async_setup()` — fail fast with clear error instead of half-supporting an impossible code path
- Remove conditional `if self.testbed_device:` guard around `_patch_device_execute_for_broker()` — now unconditional since invariant is enforced
- Consolidate `TestPatchDeviceExecuteForBroker` — all 7 tests now use the shared `_apply_broker_patch()` helper consistently

## Testing Done

### Test Commands Used

```bash
pytest tests/unit/test_ssh_base_test_validation.py -x -q  # 23 passed
pytest tests/ --collect-only  # no collection errors
pre-commit run -a  # all hooks pass
```

## Checklist

### Why this approach over the one from PR #835

PR #835 moves Genie parsing into the broker, coupling the broker to Genie and requiring a new RPC per parser method. This PR instead patches the client-side device, keeping the broker as a dumb SSH pipe while making `device.parse()`, `device.execute()`, and `device.learn()` work transparently in both modes without signature changes.

### Response to @aitestino's review findings

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | `get_cache_stats()` not thread-safe | **Fixed** — snapshot values under lock before iterating. |
| 2 | Broker-without-testbed → `AttributeError` | **Fixed** — enforced testbed invariant at the top of `_async_setup()`. Both broker and direct mode require a testbed device; the method now fails fast with a clear `ConnectionError` instead of silently deferring the crash to first command execution. |
| 3 | Dead `connection` parameter | **Fixed** — removed the parameter entirely from `_create_execute_command_method()`. It became dead after the unified `testbed_device.execute` refactor. `self.connection` attribute is retained for lifecycle management (broker cleanup branching). |
| 4 | TITLE copy-paste in `genie_parse_2` | **Fixed**. |
| 5 | `DEVICE_EXECUTE_TIMEOUT` missing from `__all__` | **Fixed**. |
| 6 | Replace `# type: ignore` with `assert ... is not None` | **Addressed differently** — used `cast()` instead. Note: recommending `assert` for type narrowing contradicts @aitestino's own [issue #679](https://github.com/netascode/nac-test/issues/679) where she stated *"using assert for type narrowing in production code is an anti-pattern"*. `cast()` has no runtime cost and is the established pattern in this codebase. |
| 7 | Unit test gap for Genie side-effects | **Fixed** — 3 tests added. |
| 8 | DRY on broker integration fixtures | **Won't fix** — nac-test's AST-based test discovery requires full inline class definitions. Extracting to base classes breaks discovery. |


aitestino (changes requested):
Hey @oboehmer, thank you for the PR — the broker architecture is solid and I really like the unified execution path eliminating mode-specific branching. The `_BrokerCliShim` approach to make Genie's supplementary `device.execute()` calls work transparently is clever.

I did a couple of reviews on this, and I think we need some changes before merge. The main themes are test coverage gaps for the new thread-safe components and a few structural items.

1. **Thread-safety test coverage for `CommandCache`** — `CommandCache` uses `threading.Lock` for all its operations, which is the right call. But there are zero concurrency tests exercising the lock under contention. From a best practices standpoint, I think we want at least one test that hammers `get`/`set` from multiple threads concurrently to prove the lock actually protects against data races. The lock is load-bearing here since multiple Genie parser threads could hit the cache simultaneously.

2. **Post-context assertions in `_apply_broker_patch` tests** — In `test_patched_execute_calls_broker`, the helper's `ExitStack` exits before the outside assertions run. The second `mock_device.execute("show version")` at line 305 hits the *cache* (populated during `call_after_patch`), not the broker path. The test passes but for the wrong reason. The `mock_future.result.assert_called_with(...)` at line 306 is valid (verifies what happened inside the context), but line 305 is testing cache behavior, not broker routing. Cleaner approach: use the `call_result` return value from `_apply_broker_patch` directly:

   ```python
   mock_device, result = self._apply_broker_patch(
       ssh_instance,
       extra_patches=[patch(self._RUN_CORO_PATH, return_value=mock_future)],
       call_after_patch=lambda dev: dev.execute("show version"),
   )
   assert result == "command output"
   mock_future.result.assert_called_with(timeout=DEVICE_EXECUTE_TIMEOUT)
   ```

3. **Test file placement** — `test_ssh_base_test_validation.py` lives at `tests/unit/` root. Looking at the existing structure, tests for `nac_test/pyats_core/common/` live in `tests/unit/pyats_core/common/` (e.g., `test_base_test_defaults.py`, `test_defaults_resolver.py`). This file should move to `tests/unit/pyats_core/common/test_ssh_base_test.py` to follow the mirror convention.

4. **Near-identical fixture templates** — `test_broker_genie_parse_1.py` and `test_broker_genie_parse_2.py` in `tests/integration/fixtures/templates_broker/tests/` are identical except for the TITLE suffix `(1)`/`(2)` and the class name. All module constants (`DESCRIPTION`, `SETUP`, `PROCEDURE`, `PASS_FAIL_CRITERIA`, `TEST_CONFIG`) and the entire `verify_item` method are duplicated verbatim. I understand these test caching behavior (file 1 = misses, file 2 = hits), but the DRY violation is significant. Consider extracting a shared base class or a parametrized template that only varies the title/class name.

5. **`DEVICE_EXECUTE_TIMEOUT` should be env-configurable** — This constant is hardcoded as `120` in `constants.py`, but the same file uses `get_positive_numeric_env()` for `PYATS_OUTPUT_BUFFER_LIMIT`, `PIPE_DRAIN_DELAY_SECONDS`, `BATCH_SIZE`, etc. A 120-second timeout for broker-routed commands is a reasonable default, but some environments (slow WAN links, large outputs) may need to tune it. Suggested approach:

   ```python
   DEVICE_EXECUTE_TIMEOUT: int = get_positive_numeric_env(
       "NAC_TEST_DEVICE_EXECUTE_TIMEOUT", 120, int
   )
   ```

What do you think?

oboehmer:
Ciao @aitestino .. thanks for the review, I addressed all  (via a73f6a2) but item 4, which I would still push back on: The test fixture files (test_broker_genie_parse_1.py and test_broker_genie_parse_2.py) are processed in a full nac-test integration test run, and thus are subject to our test type discovery which AST-parses the files and looks for classes, aetest decorators and the like. Subclassing will break this, and I feel integration and e2e tests should process "real" test files. I acknowledge that there is a lot of duplication of test case fixtures in tests/integration and tests/e2e, but I feel this is acceptable and meets the test objective of processing "real" test cases. 

please re-review.

aitestino (changes requested):
Hey @oboehmer, thanks for the fast turnaround on the last review — commit a73f6a2 lands cleanly on items 1, 2, 3, and 5.

**Items worth addressing before merge:**

1. **`test_command_cache.py` placement contradicts the mirroring rationale from the same commit** — the file is at `tests/unit/pyats_core/common/test_command_cache.py`, but its source lives at `nac_test/pyats_core/ssh/command_cache.py`. `tests/unit/pyats_core/ssh/` doesn't exist yet. Should be `tests/unit/pyats_core/ssh/test_command_cache.py` (create the dir + `__init__.py`) — same reasoning you used to move `test_ssh_base_test.py`.

2. **`broker_execute` timeout failure path is untested** — `future.result(timeout=DEVICE_EXECUTE_TIMEOUT)` will raise `TimeoutError` if the broker hangs. The two tests at `test_ssh_base_test.py:306, 367` assert the timeout arg is passed (`mock_future.result.assert_called_with(timeout=…)`) but never side-effect a `TimeoutError` to verify it propagates and that the failed command is NOT cached. One extra test would close it.

3. **Concurrency tests don't force the compound-sequence race** — your `TestCommandCacheConcurrency` docstring correctly notes: *"The lock is still correct and necessary for the compound check-read-delete sequence in get() and for forward compatibility."* That's exactly right. But the two tests as written don't actually exercise the compound path (both use `ttl=3600`, so `get()` never hits the expiry-delete branch). A third test with `ttl=0` (or a very short TTL) plus concurrent get/set on the same key would actually verify the lock catches removal. Cheap addition, buys real regression coverage under both CPython and free-threaded Python.

4. **New monkey-patching interacts with a pre-existing loose pyats pin** — `pyproject.toml:47` pins `pyats>=26.5` with no upper bound (this is pre-existing, not your change). `_BrokerCliShim` only implements `.execute()` — a future Genie release touching `device.cli.timeout`, `.log`, or any other attribute would throw `AttributeError` deep in Genie internals. Two suggestions, either in this PR or a follow-up: (a) tighten to `>=26.5,<27`, (b) add a defensive `__getattr__` on `_BrokerCliShim` that logs and raises a clear framework error rather than a bare AttributeError.

**Smaller nits:**

5. **`broker_execute` docstring missing Args/Returns** — `ssh_base_test.py:313-321`. The sibling `execute_command` closure has them; matching keeps the style consistent.

6. **`test_patched_execute_calls_broker` tautology** — `test_ssh_base_test.py:305`, `assert result == "command output"` where `mock_future.result = Mock(return_value="command output")`. The Mock returns what it was told to return, so that assertion doesn't test production behavior. The `mock_future.result.assert_called_with(timeout=DEVICE_EXECUTE_TIMEOUT)` on the next line IS meaningful. Cleanest strengthening: patch `broker_client.execute_command` directly (not `run_coroutine_threadsafe`) so the coroutine target is actually asserted.

7. **`## Fixes` CHANGELOG header is novel** — the rest of CHANGELOG uses either `## Other Improvements` (2.0.0 style) or flat bullets (pre-2.0.0). Either is fine; `## Fixes` doesn't match. Style-only.

What do you think?

oboehmer:
Thanks for the additional review, @aitestino .. Addressed them, please re-review

aitestino (approved):
Hey @oboehmer, thanks for the documentation addition in `dd86c53` — the thread-hop trade-off comment at the `run_in_executor` call site covers all three points cleanly (µs vs ms overhead, parse_output requiring run_in_executor anyway, unified path eliminating mode-dependent bugs). This was the only blocking item from my third review.

The follow-up issues (#920–#927) are tracked separately and none block the merge. Approving.

PR #874, opened 2026-06-19
Update `iterate_list_chunked` to support dict as parent_key

rchrabas:
## Description

Framework allows the following structure to split test into multiple files

{# iterate_list_chunked <YAML_PATH_TO_LIST> <LIST_ITEM_ID> <JINJA_VARIABLE_NAME> <OBJECT_PATH> <CHUNK_SIZE> #}

<OBJECT_PATH> and have parent_key.child_key, however it assumes that parent_key is a list.
This PR allows parent_key to be a dict.

## Closes
- #875 

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Refactoring / Technical debt (internal improvements with no user-facing changes)
- [ ] Documentation update

## Test Framework Affected

- [ ] PyATS
- [X] Robot Framework
- [ ] Both

- [ ] ACI (APIC)
- [ ] Meraki (Cloud-managed)
- [ ] Hyperfabric
- [X] All architectures

## Platform Tested

- [X] macOS (version tested: 26.5.1)

## Key Changes

Summarize the main changes in this PR:
- Update RobotWriter._chunk_nested_objects()
- Add tests

## Testing Done

- [X] Unit tests added/updated
- [ ] Integration tests performed
  - [X] Robot Framework tests executed successfully
- [X] All existing tests pass (`pytest` / `pre-commit run -a`)

### Test Commands Used

```bash
# Commands used to test changes
```

## Checklist

- [X] Code follows project style guidelines (`pre-commit run -a` passes)
- [X] Self-review of code completed
- [X] Code is commented where necessary (especially complex logic)
- [X] No new warnings introduced

### DIFF
--- a/nac_test/robot/robot_writer.py
+++ b/nac_test/robot/robot_writer.py
@@ -189,6 +189,22 @@ class RobotWriter:
         elif len(path_parts) == 2:
             # Nested case: collect objects from nested structure
             parent_key, child_key = path_parts
+            parent = data.get(parent_key)
+
+            # Check if parent is a dict
+            if isinstance(parent, dict):
+                objects = parent.get(child_key, [])
+                if not isinstance(objects, list):
+                    return [data]
+                chunked_data = []
+                for i in range(0, len(objects), chunk_size):
+                    chunked_item = copy.deepcopy(data)
+                    chunked_item[parent_key][child_key] = objects[i : i + chunk_size]
+                    chunked_data.append(chunked_item)
+
+                return chunked_data
+
+            # Continue with assumption that parent is a list
             all_objects = []
 
             # Collect all nested objects with their parent context
--- a/nac_test/robot/robot_writer.py
+++ b/nac_test/robot/robot_writer.py
@@ -170,11 +170,13 @@ class RobotWriter:
             List of modified data structures, each containing a subset of
             objects, or ``[data]`` unchanged if the path does not hold a list
         """
-        container = data
+        container: Any = data
         for key in path_parts[:-1]:
-            container = container[key]
-        objects = container.get(path_parts[-1], [])
-        if not isinstance(objects, list):
+            container = container.get(key)
+            if not isinstance(container, dict):
+                return [data]
+        objects = container.get(path_parts[-1])
+        if not isinstance(objects, list) or not objects:
             return [data]
 
         chunked_data = []
--- a/nac_test/robot/robot_writer.py
+++ b/nac_test/robot/robot_writer.py
@@ -260,7 +260,7 @@ class RobotWriter:
                     chunked_data.append(chunked_item)
 
                 return chunked_data
-            
+
             else:
                 raise ValueError(
                     f"Parent key '{parent_key}' is neither a list nor a dict in the data structure."
--- a/nac_test/robot/robot_writer.py
+++ b/nac_test/robot/robot_writer.py
@@ -150,6 +150,44 @@ class RobotWriter:
                 return Path(*paths[:-1], "_" + paths[-1])
         return Path(os.path.join(*paths))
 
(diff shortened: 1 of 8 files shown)


oboehmer:
Thanks for the contribution, Rafal! Would you mind creating an issue describing the use case for this dict chunking, like which which architecture/test case benefits from this enhancement? Will review later this week

oboehmer (changes requested):
Thanks for the contribution and the thorough test coverage, Rafal — the mutation check and sibling-preservation assertions are particularly nice, and the use case in #875 makes the motivation clear.

I would like to ask for a few changes before this is ready to merge:

**Code structure**

The new block is an early-return island inserted inside the existing `len(path_parts) == 2` branch, which already handles a distinct algorithm (flattening child objects across a list of parents). Since the two cases — dict parent and list parent — produce fundamentally different results, making them explicit siblings rather than one being an early exit inside the other would make the intent clearer.

It's also worth noticing that the dict-parent logic is structurally identical to the `len(path_parts) == 1` block: both navigate to a list at a known path and chunk it in-place. A small private helper could remove that duplication and serve both callers.

One shadowing nit: `parent = data.get(parent_key)` is introduced at the top of the branch, but the existing list-parent loop below does `for parent in data.get(parent_key, []):` — the loop variable silently shadows it.

**Test coverage**

Looking at the overall coverage picture:

| Path | Unit tested | Integration tested |
|---|---|---|
| 1-level (`hosts`) | No | Yes (existing fixture) |
| 2-level, list parent (`services.endpoints`) | Yes (this PR) | Yes (existing fixture) |
| 2-level, dict parent (`objects.hosts`) | Yes (this PR) | No |

The PR leaves two gaps (one is pre-existing, but as you touched this area ;-) ): the 1-level path has no unit test, and the new dict-parent path has no integration test.
Could I ask you to close these gaps: 
- add a 1-level case to `TestChunkNestedObjects` — straightforward given the tests already there:
- add a fixture and integration test to `test_cli_rendering.py` covering the dict-parent path, using a data model like the FMC shape from #875 (a dict parent with a `hosts` list underneath).

That gives symmetric end-to-end coverage across all three paths.

**Documentation**

Could you please add a short chapter to the "Rendering Directives" chapter in `README.md` so users are aware of this new capability.


Please let me know if you have questions or concerns about any of the above, and thanks again for raising this.

danischm (approved):

nac_test/robot/robot_writer.py
```diff
@@ -150,6 +150,44 @@ def _fix_duplicate_path(self, *paths: str) -> Path:
...
+            path_parts: Keys navigating from ``data`` to the list to chunk
+            chunk_size: Number of objects per chunk
+
+        Returns:
+            List of modified data structures, each containing a subset of
+            objects, or ``[data]`` unchanged if the path does not hold a list
+        """
+        container = data
+        for key in path_parts[:-1]:
+            container = container[key]
```
oboehmer: I feel we should guard this against the key not present? We can't fully control the input side, so maybe an optional key might not be present? All other scenarios use `.get()`, so we should possibly do too?

```suggestion
            container = container.get(key)
            if not isinstance(container, dict):
                return [data]
```

oboehmer (changes requested):
thanks for addressing all the changes. I noticed one are where we could harden the code a bit more in case the user's input isn't well-formed. 

possible trigger: a user writes `{# iterate_list_chunked fmc.domains name domain_name objects.hosts 2 #}` but one of their domain items simply doesn't have an objects key (missing/optional field in the data model). We would crash in robot_writer with no pointer at the user's data or template.

Does this make sense?

oboehmer (approved):

